PR #6, opened 2020-05-19
Add viewer tool

s9latimm:
M	CMakeLists.txt
M	tools/fm-viewer/FeatureModelViewer.cpp

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ jobs:
             - g++-9
             - clang-tidy-10
       env: CXX=g++-9 BUILD_TYPE=Debug
-    - name: GCC-9 Release\
+    - name: GCC-9 Release
       stage: Latest
       addons: *gcc9
       env: CXX=g++-9 BUILD_TYPE=Release
@@ -81,7 +81,7 @@ script:
   - cmake --build . --target tidy-vara-feature
   - cmake --build . --target check-vara-feature-python
   - cmake --build . --target check-vara-fm-viewer
-  -
+
 branches:
   only:
     - vara
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os: linux
 dist: bionic
 
 before_install:
-  - sudo apt-get -y install libstdc++6 libxml2-dev
+  - sudo apt-get -y install libstdc++6 libxml2-dev graphviz
 
 jobs:
   include:
@@ -67,7 +67,7 @@ jobs:
             - g++-9
             - clang-tidy-10
       env: CXX=g++-9 BUILD_TYPE=Debug
-    - name: GCC-9 Release
+    - name: GCC-9 Release\
       stage: Latest
       addons: *gcc9
       env: CXX=g++-9 BUILD_TYPE=Release
@@ -80,7 +80,8 @@ script:
   - cmake --build . --target check-vara-feature-unittests
   - cmake --build . --target tidy-vara-feature
   - cmake --build . --target check-vara-feature-python
-
+  - cmake --build . --target check-vara-fm-viewer
+  -
 branches:
   only:
     - vara
--- a/include/vara/Feature/FeatureModel.h
+++ b/include/vara/Feature/FeatureModel.h
@@ -159,40 +159,44 @@ struct DOTGraphTraits<vara::feature::FeatureModel *>
   }
 
   static std::string getGraphProperties(const vara::feature::FeatureModel *FM) {
-    return "\tgraph[pad=0.5,nodesep=2,ranksep=2,splines=true,newrank=true]";
+    return "\tgraph[pad=0.5,nodesep=2,ranksep=2,splines=true,newrank=true,"
+           "fontname=\"CMU Typewriter\"]";
   }
 
   static std::string
   getEdgeAttributes(const vara::feature::Feature *Node,
                     vara::feature::Feature::feature_iterator I,
                     const vara::feature::FeatureModel *FM) {
-    std::stringstream S;
-    S << "arrowhead=\"";
-    if ((*I)->isOptional()) {
-      S << 'o';
-    }
-    S << "dot\" ";
-    return S.str();
+    return std::string("arrowhead=")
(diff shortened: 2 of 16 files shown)


include/vara/Feature/FeatureModel.h
```diff
@@ -191,6 +175,26 @@ struct DOTGraphTraits<vara::feature::FeatureModel *>
     return S.str();
   }
 
+  static void buildCluster(llvm::raw_ostream &O, vara::feature::Feature *Node,
+                           int L = 1) {
```
vulder: what is `L` could you please add a comment for the function that also describes the parameters

tools/fm-viewer/FeatureModelViewer.cpp
```diff
@@ -1,20 +1,81 @@
...
+                         << "\n";
+          }
+        }
+        return 0;
+      }
+    } else {
+      llvm::errs() << "error: Could not open file.\n";
     }
+  } else {
+    llvm::errs() << "error: Expected single file.\n";
```
vulder: We should restructure this to make error checking and return and not nest everything because this becomes hard to reason about. For example:
```
if (!FM) {
  llvm::errs() << "Foo not feature model\n";
  return 1;
}
... continue the good case
```
Otherwise, we create on nesting depth per error case.

tools/fm-viewer/FeatureModelViewer.cpp
```diff
@@ -1,20 +1,81 @@
...
+
+static llvm::cl::list<std::string> FileNames(llvm::cl::Positional,
+                                             llvm::cl::desc("<file>"),
+                                             llvm::cl::cat(FMViewerCategory));
+
+int main(int Argc, char **Argv) {
+  llvm::InitLLVM X(Argc, Argv);
+
+  llvm::cl::HideUnrelatedOptions(FMViewerCategory);
+
```
vulder, line 43:
```suggestion
```

tools/fm-viewer/FeatureModelViewer.cpp
```diff
@@ -1,20 +1,81 @@
...
+  llvm::cl::ParseCommandLineOptions(Argc, Argv,
+                                    "Shows feature model as graph.");
```
vulder:
```suggestion
  llvm::cl::ParseCommandLineOptions(Argc, Argv,
                                    Overview, /*Errs=*/nullptr, FlagsEnvVar);
```

tools/fm-viewer/FeatureModelViewer.cpp
```diff
@@ -1,20 +1,81 @@
...
+    Dump("dump", llvm::cl::desc("Dump feature model to stdout and exit."),
+         llvm::cl::init(false), llvm::cl::cat(FMViewerCategory));
+
+static llvm::cl::list<std::string> FileNames(llvm::cl::Positional,
+                                             llvm::cl::desc("<file>"),
+                                             llvm::cl::cat(FMViewerCategory));
+
+int main(int Argc, char **Argv) {
+  llvm::InitLLVM X(Argc, Argv);
+
```
vulder:
```suggestion
  const char *FlagsEnvVar = "FM_VIEWER_FLAGS";
  const char *Overview = R"(Please add overview/intro text here)"
```

tools/fm-viewer/FeatureModelViewer.cpp
```diff
@@ -1,20 +1,81 @@
...
+    std::ifstream DocFile(FileNames[0]);
     if (DocFile.is_open()) {
-      std::string DocRaw((std::istreambuf_iterator<char>(DocFile)),
-                         std::istreambuf_iterator<char>());
+      std::unique_ptr<vara::feature::FeatureModel> FM;
+
+      if (Xml) {
+        std::string DocRaw((std::istreambuf_iterator<char>(DocFile)),
+                           std::istreambuf_iterator<char>());
+        FM = vara::feature::FeatureModelXmlParser(DocRaw).buildFeatureModel();
```
vulder:
```suggestion
        FM = vara::feature::FeatureModelXmlParser(std::move(DocRaw)).buildFeatureModel();
```

tools/fm-viewer/FeatureModelViewer.cpp
```diff
@@ -1,20 +1,81 @@
...
-      llvm::ViewGraph(FM.get(), llvm::Twine(FM->getName()));
-      return 0;
+
+      if (FM) {
+        if (Dump) {
+          FM->dump();
+        } else {
+          std::string Filename =
+              llvm::WriteGraph(FM.get(), llvm::Twine(FM->getName()));
+          if (!Filename.empty()) {
```
vulder: Is empty a usual thing that can happen here? If not I would suggest adding and assert and `return 1` with an error message;

vulder (changes requested):

tools/fm-viewer/FeatureModelViewer.cpp
s9latimm: Returning without error mimics the behavior of `llvm::ViewGraph` (`GraphWriter.h`) given `WriteGraph` failed to open a file for writing (which may be kind of usual). So instead, I would terminate with an error - as the tool basically failed.

.travis.yml
```diff
@@ -4,7 +4,7 @@ os: linux
 dist: bionic
 
 before_install:
-  - sudo apt-get -y install libstdc++6 libxml2-dev
+  - sudo apt-get -y install libstdc++6 libxml2-dev graphviz
```
vulder, line 7: we should add this dep also to the readme

.travis.yml
```diff
@@ -67,7 +67,7 @@ jobs:
             - g++-9
             - clang-tidy-10
       env: CXX=g++-9 BUILD_TYPE=Debug
-    - name: GCC-9 Release
+    - name: GCC-9 Release\
```
vulder: What is the meaning of this `\`?

vulder (approved):

PR #7, opened 2020-05-22
Move namespace

s9latimm:


include/vara/Feature/Feature.h
```diff
@@ -27,36 +27,53 @@ class Feature {
   using relationship_iterator = RelationshipListTy ::iterator;
 
 private:
-  string Name;
+  std::string Name;
+  llvm::SetVector<llvm::Value *> Values;
+  size_t ID;
+  bool Opt;
```
vulder, line 33: should be at the end of the variables

include/vara/Feature/Feature.h
```diff
@@ -27,36 +27,53 @@ class Feature {
   using relationship_iterator = RelationshipListTy ::iterator;
 
 private:
-  string Name;
+  std::string Name;
+  llvm::SetVector<llvm::Value *> Values;
+  size_t ID;
```
vulder, line 32: I think we can remove these two but this also requires changes on the VaRA side.

include/vara/Feature/Feature.h
```diff
@@ -27,36 +27,53 @@ class Feature {
...
+      : Name(Name), ID(std::hash<std::string>{}(Name)), Opt(false) {
+    Values.insert(Val);
+  }
+
   Feature(const Feature &) = delete;
   Feature &operator=(const Feature &) = delete;
   virtual ~Feature() = default;
 
-  [[nodiscard]] string getName() const { return Name; }
+  [[nodiscard]] std::string getName() const { return Name; }
```
vulder, line 57: Could be a string ref

include/vara/Feature/Feature.h
```diff
@@ -27,36 +27,53 @@ class Feature {
...
-  void print(std::ostream &Out) const { Out << toString() << std::endl; }
-  void print(llvm::raw_ostream &Out) const { Out << toString() << '\n'; }
+  void print(llvm::raw_ostream &OS) const {
+    OS << getName() << " (";
+    for (auto *Val : Values) {
+      OS << Val << " ";
+    }
+    OS << ")"
+       << " ID: " << getID();
+  }
```
vulder, line 70: If we can remove ID and values we can go back to the toString option, which would also be better because we allow subclasses to override `toString`.

vulder (changes requested):

include/vara/Feature/Feature.h
s9latimm: Will be resolved with #6 and merged afterwards

PR #10, opened 2020-05-23
Implements a basic set of python bindings.

vulder:
Adds python bindings for the essential functions.
Adds support for:
- working with features
- location integration

### DIFF
--- a/bindings/python/vara-feature/pybind_Feature.cpp
+++ b/bindings/python/vara-feature/pybind_Feature.cpp
@@ -44,15 +44,63 @@ void init_feature_module_feature(py::module &M) {
       .def("add_parent", &vf::Feature::addParent,
            R"pbdoc(Add a parent feature to this feature.)pbdoc")
       .def("is_parent", &vf::Feature::isParent,
-           R"pbdoc(Checks if a Feature is a parent of this one.)pbdoc");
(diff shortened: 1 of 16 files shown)


bindings/python/vara-feature/pybind_Feature.cpp
```diff
@@ -9,24 +10,112 @@
...
-           R"pbdoc(`True` if the feature is optional.)pbdoc");
+           R"pbdoc(`True` if the feature is optional.)pbdoc")
+      .def(
+          "is_root", &vf::Feature::isRoot,
+          R"pbdoc(`True` if this is the root feature of a `FeatureModel`.)pbdoc")
+      //===----------------------------------------------------------------===//
+      // Children
+      .def(
+          "children",
+          [](vf::Feature &F) { return py::make_iterator(F.begin(), F.end()); },
```
padupr, line 26: Iterators are not indexable, I don't think I need it but it's not intuitive.

vulder: As we talk today, iterators are the better interface for the Cpp side but in python `list()` conversions work if one wants to use indexing.
It looks like this is the more pythonic way :laughing:

padupr (approved):

PR #14, opened 2020-06-12
Adds a FeatureModelBuilder and iterates depth first

s9latimm:
- [x] FeatureModelBuilder
- [x] depth first iterator
- [x] comments
- [x] update bindings
- [x] clang-tidy


include/vara/Feature/Feature.h
```diff
@@ -1,13 +1,15 @@
...
-#include "vara/Feature/Location.h"
-#include "vara/Feature/Relationship.h"
+#include "vara/Feature/FeatureSourceRange.h"
 
-#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/DenseSet.h"
 #include "llvm/ADT/iterator_range.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <set>
```
vulder, line 10: seems like this is not used

include/vara/Feature/Feature.h
```diff
@@ -20,28 +22,37 @@ namespace vara::feature {
...
+  void addAlternative(Feature *F) { Alternatives.insert(F); }
+
+  void addImplication(Feature *F) { Implications.insert(F); }
 
 protected:
-  Feature(string Name, bool Opt, std::optional<Location> Loc)
-      : Name(std::move(Name)), Opt(Opt), Loc(std::move(Loc)) {}
+  Feature(string Name, bool Opt, std::optional<FeatureSourceRange> Loc,
+          Feature *Parent)
+      : Name(std::move(Name)), Opt(Opt), Loc(std::move(Loc)), Parent(Parent) {}
```
vulder: Protected section should come after public section

include/vara/Feature/Feature.h
```diff
@@ -20,28 +22,37 @@ namespace vara::feature {
...
+
+  void addChild(Feature *F) { Children.insert(F); }
+
+  void setParent(Feature *F) { Parent = F; }
+
+  void addExclude(Feature *F) { Excludes.insert(F); }
+
+  void addAlternative(Feature *F) { Alternatives.insert(F); }
+
+  void addImplication(Feature *F) { Implications.insert(F); }
```
vulder: Private sections should come after public/protected sections

include/vara/Feature/Feature.h
```diff
@@ -52,7 +63,11 @@ class Feature {
 
   [[nodiscard]] bool isOptional() const { return Opt; }
 
-  [[nodiscard]] bool isRoot() const { return Parents.empty(); }
+  [[nodiscard]] virtual bool isBinary() const = 0;
+
+  [[nodiscard]] virtual bool isNumeric() const = 0;
```
vulder: Why to we implement specific check functions when we use inheritance? If we need this information we should impl llvm style type info with kind enums. Then we can write `llvm::isa<BinaryFeature *>()`

include/vara/Feature/Feature.h
```diff
@@ -203,30 +169,41 @@ class Feature {
...
-           llvm::StringRef(F.getName()).lower();
+  bool operator==(const vara::feature::Feature &F) const {
+    return getName().lower() == F.getName().lower();
   }
 
-  bool operator<(const Feature &F) const {
-    return llvm::StringRef(this->getName()).lower() <
-           llvm::StringRef(F.getName()).lower();
+  struct FeatureDepthFirstComparator {
+    bool operator()(vara::feature::Feature *, vara::feature::Feature *) const;
```
vulder: Is a ptr here required? I would assume `const &` should be enough here

include/vara/Feature/FeatureModel.h
```diff
@@ -17,7 +19,8 @@ namespace vara::feature {
 class FeatureModel {
 public:
   using FeatureMapTy = llvm::StringMap<std::unique_ptr<Feature>>;
-  using ConstraintsTy = std::vector<std::vector<std::pair<Feature *, bool>>>;
+  using ConstraintTy = llvm::SmallVector<std::pair<std::string, bool>, 3>;
+  using ConstraintsTy = std::vector<ConstraintTy>;
 
 private:
```
vulder: same public/protected/private order problem

include/vara/Feature/FeatureModel.h
```diff
@@ -68,29 +69,135 @@ class FeatureModel {
     }
   };
 
-  FeatureModelIter begin() { return FeatureModelIter(Features.begin()); }
+public:
+  using feature_model_iterator = FeatureModelIter;
+  using const_feature_model_iterator = FeatureModelIter;
+
+  feature_model_iterator begin() { return FeatureModelIter(Features.begin()); }
```
vulder: Why is the FeatureModelIter private but exposed throught the public interface?

include/vara/Feature/FeatureModel.h
```diff
@@ -68,29 +69,135 @@ class FeatureModel {
...
+
+  [[nodiscard]] const_feature_model_iterator begin() const {
+    return FeatureModelIter(Features.begin());
+  }
 
   FeatureModelIter end() { return FeatureModelIter(Features.end()); }
 
+  [[nodiscard]] const_feature_model_iterator end() const {
+    return FeatureModelIter(Features.end());
+  }
```
vulder, line 81: Please add also add llvm style range function `features`

include/vara/Feature/FeatureModel.h
```diff
@@ -68,29 +69,135 @@ class FeatureModel {
...
-    for (const auto &F : this->Features) {
-      F.second->dump();
-      llvm::outs() << "\n";
+  void dump() const;
+
+  class FeatureModelBuilder {
+    using EdgeMapType =
+        typename llvm::StringMap<llvm::SmallVector<std::string, 3>>;
+
+    std::string VmName;
```
vulder: private/public order

include/vara/Feature/FeatureModel.h
```diff
@@ -68,29 +69,135 @@ class FeatureModel {
...
+          .try_emplace(
+              Key, std::make_unique<BinaryFeature>(Key, Opt, std::move(Loc)))
+          .second;
+    }
+
+    FeatureModelBuilder *addChild(const std::string &P, const std::string &C) {
+      Children[P].push_back(C);
+      Parents[C] = P;
+      return this;
+    }
```
vulder: Does it make sense to also add overloads for when we have a `Feature &` and what to add a child etc.?

include/vara/Feature/FeatureModelParser.h
```diff
@@ -99,28 +99,15 @@ class FeatureModelXmlParser : public FeatureModelParser {
 
 private:
```
vulder: This private section should also be at the bottom

include/vara/Feature/FeatureSort.h
```diff
@@ -0,0 +1,40 @@
+#ifndef VARA_FEATURE_FEATURESORT_H
+#define VARA_FEATURE_FEATURESORT_H
+
+#include "FeatureModel.h"
+
+namespace vara::feature {
+
+class FeatureSort {
```
vulder: Sounds like an algorithm. I would prefer name like `OrderedFeatureVector`

include/vara/Feature/FeatureSort.h
```diff
@@ -0,0 +1,40 @@
...
+public:
+  using feature_heap_iterator = typename std::vector<Feature *>::iterator;
+  using const_feature_heap_iterator =
+      typename std::vector<Feature *>::const_iterator;
+
+private:
```
vulder: private to the bottom

include/vara/Feature/FeatureSort.h
```diff
@@ -0,0 +1,40 @@
...
+  std::vector<Feature *> Features;
+
+public:
+  template <typename T> FeatureSort(T FM) {
+    for (auto F = FM->begin(); F != FM->end(); ++F) {
+      Features.push_back(*F);
+    }
+    std::sort(Features.begin(), Features.end(),
+              Feature::FeatureDepthFirstComparator());
+  }
```
vulder: Currently, this is not a heap. Didn't you want to make it a heap?

include/vara/Feature/FeatureSort.h
vulder: Currently, there are insertion methods missing. And I also do not see how this is used in the FM. I'll let you finish the impl before I'll continue to look at stuff :)

vulder (changes requested):
sry, for the public/protected/private stuff I missed this last time

include/vara/Feature/FeatureSort.h
s9latimm: Building a heap I would need to call `sort_heap` to receive a sorted collection while losing the heap property. As a custom iterator does not seem reasonable at this point, I used a  comparator with insertion at `std::upper_bound` instead.

bindings/python/tests/feature/test_feature.py
```diff
@@ -28,74 +29,18 @@ def test_is_root(self):
...
-    def test_add_alternative(self):
-        """ Checks if a feature is correctly added as an alternative from
-        another. """
-        root_feature = feature.BinaryFeature("root", False)
-        test_feature = feature.BinaryFeature("Test", False)
-
-        test_feature.add_alternative(root_feature)
-
-        self.assertTrue(test_feature.is_alternative(root_feature))
+    # def test_iter_children(self):
```
vulder: why is this case outcommented?

bindings/python/vara-feature/pybind_FeatureSourceRange.cpp
```diff
@@ -0,0 +1,70 @@
...
+namespace fs = std::experimental::filesystem;
+#else
+#include <filesystem>
+namespace fs = std::filesystem;
+#endif
+
+namespace vf = vara::feature;
+namespace py = pybind11;
+
+void init_feature_module_location(py::module &M) {
```
vulder: `init_feature_location_module`

bindings/python/vara-feature/pybind_FeatureSourceRange.cpp
```diff
@@ -0,0 +1,70 @@
...
+      .def_property_readonly(
+          "path",
+          [](vf::FeatureSourceRange &Loc) { return Loc.getPath().string(); },
+          R"pbdoc(Path to the source file)pbdoc")
+      .def("get_start", &vf::FeatureSourceRange::getStart,
+           py::return_value_policy::reference,
+           R"pbdoc(Get the start `LineColumnOffset` of this `Location`.)pbdoc")
+      .def("get_end", &vf::FeatureSourceRange::getEnd,
+           py::return_value_policy::reference,
+           R"pbdoc(Get the end `LineColumnOffset` of this `Location`.)pbdoc")
```
vulder: If I remember correctly pascal migrated accessors to python properties, we should use this here too.

include/vara/Feature/Feature.h
```diff
@@ -21,44 +22,25 @@ namespace vara::feature {
...
-  FeatureListType Children;
-  FeatureListType Excludes;
-  FeatureListType Implications;
-  FeatureListType Alternatives;
-  RelationshipListTy Relationships;
-
-protected:
-  Feature(string Name, bool Opt, std::optional<Location> Loc)
-      : Name(std::move(Name)), Opt(Opt), Loc(std::move(Loc)) {}
+  enum FeatureType { BINARY, NUMERIC };
```
```suggestion
  enum FeatureKind { FK_Binary, FK_Numeric };
```

include/vara/Feature/Feature.h
```diff
@@ -21,44 +22,25 @@ namespace vara::feature {
...
 
-public:
   Feature(const Feature &) = delete;
-  Feature &operator=(const Feature &) = delete;
   virtual ~Feature() = default;
 
+  [[nodiscard]] FeatureType getType() const { return T; }
```
```suggestion
  [[nodiscard]] FeatureKind getKind() const { return T; }
```

include/vara/Feature/Feature.h
```diff
@@ -203,34 +136,90 @@ class Feature {
   }
   [[nodiscard]] const_feature_iterator end() const { return children_end(); }
 
-  [[nodiscard]] Location* getLocation() {
+  //===--------------------------------------------------------------------===//
+  // Operators
+  Feature &operator=(const Feature &) = delete;
```
vulder: Copy assignment operator should be moved to the ctor if possible.

include/vara/Feature/Feature.h
```diff
@@ -203,34 +136,90 @@ class Feature {
...
 
-  bool operator<(const Feature &F) const {
-    return llvm::StringRef(this->getName()).lower() <
-           llvm::StringRef(F.getName()).lower();
-  }
+  LLVM_DUMP_METHOD
+  void dump() const { llvm::outs() << toString() << "\n"; }
+
+protected:
+  Feature(FeatureType T, string Name, bool Opt,
```
```suggestion
  Feature(FeatureKind Kind, string Name, bool Opt,
```

include/vara/Feature/Feature.h
```diff
@@ -203,34 +136,90 @@ class Feature {
...
+
+  void addChild(Feature *F) { Children.insert(F); }
+
+  void setParent(Feature *F) { Parent = F; }
+
+  void addExclude(Feature *F) { Excludes.insert(F); }
+
+  void addAlternative(Feature *F) { Alternatives.insert(F); }
+
+  void addImplication(Feature *F) { Implications.insert(F); }
```
```suggestion
  void addChild(Feature *F) { Children.insert(F); }

  void setParent(Feature *F) { Parent = F; }

  void addExclude(Feature *F) { Excludes.insert(F); }

  void addAlternative(Feature *F) { Alternatives.insert(F); }

  void addImplication(Feature *F) { Implications.insert(F); }
  
  FeatureType T;
  string Name;
  bool Opt;
  std::optional<FeatureSourceRange> Loc;
  Feature *Parent;
  FeatureSetType Children;
  FeatureSetType Excludes;
  FeatureSetType Implications;
  FeatureSetType Alternatives;
```
Nit: I would prefer to have all members at the end of the class.

include/vara/Feature/Feature.h
```diff
@@ -203,34 +136,90 @@ class Feature {
...
-                std::optional<Location> Loc = std::nullopt)
-      : Feature(std::move(Name), Opt, std::move(Loc)) {}
+  BinaryFeature(string Name, bool Opt = false,
+                std::optional<FeatureSourceRange> Loc = std::nullopt,
+                Feature *Parent = nullptr, FeatureSetType Children = {},
+                FeatureSetType Excludes = {}, FeatureSetType Implications = {},
+                FeatureSetType Alternatives = {})
+      : Feature(BINARY, std::move(Name), Opt, std::move(Loc), Parent,
+                std::move(Children), std::move(Excludes),
+                std::move(Implications), std::move(Alternatives)) {}
```
vulder: Does it make sense to make the ctor protected and only offer a factory function on the `FeatureModel`? Otherwise, we cannot guarantee that the values set in the ctor are consistent with the rest of the model.

include/vara/Feature/FeatureModel.h
```diff
@@ -1,37 +1,40 @@
...
-      : Name(std::move(Name)), RootPath(std::move(RootPath)),
-        Features(std::move(Features)), Constraints(std::move(Constraints)),
-        Root(this->Features["root"].get()) {}
+               Feature *Root)
+      : Name(std::move(Name)), RootPath(std::move(RootPath)), Root(Root),
+        Features(std::move(Features)) {
+    // Insert all values into ordered data structure.
+    for (const auto &KV : this->Features) {
+      OrderedFeatures.insert(KV.getValue().get());
+    }
```
vulder, line 34: Update to new insert method. (see below)

include/vara/Feature/FeatureSourceRange.h
```diff
@@ -0,0 +1,102 @@
...
+//===----------------------------------------------------------------------===//
+
+class FeatureSourceRange {
+public:
+  class FeatureSourceLocation {
+
+  public:
+    FeatureSourceLocation(int Line, int Column) : Line(Line), Column(Column) {}
+
+    void setLineNumber(int L) { this->Line = L; }
```
vulder:
```suggestion
    void setLineNumber(int LineNumber) { this->Line = LineNumber; }
```

include/vara/Feature/FeatureSourceRange.h
```diff
@@ -0,0 +1,102 @@
...
+    [[nodiscard]] int getLineNumber() const { return this->Line; }
+
+    void setColumnOffset(int C) { this->Column = C; }
```
vulder:
```suggestion
    void setColumnOffset(int ColumnOffset) { this->Column = ColumnOffset; }
```

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -0,0 +1,46 @@
...
+  OrderedFeatureVector() = default;
+
+  /// Insert feature while preserving ordering.
+  void insert(Feature *F) {
+    Features.insert(
+        std::upper_bound(Features.begin(), Features.end(), F,
+                         [](vara::feature::Feature *A,
+                            vara::feature::Feature *B) { return *A < *B; }),
+        F);
+  }
```
vulder, line 34: It would be helpful to provide different overloads for insert, especially, the one that takes and iterator range :wink: .

lib/Feature/Feature.cpp
```diff
@@ -1,57 +1,34 @@
 #include "vara/Feature/Feature.h"
 
+#include <experimental/iterator>
```
vulder:
```suggestion
#include <iterator>
```
Why would we need experimental stuff here?

unittests/Feature/Feature.cpp
```diff
@@ -1,59 +1,79 @@
...
+  auto FM = FeatureModelBuilder().buildSimpleFeatureModel(
+      {{"a", "aa"}, {"root", "b"}, {"root", "a"}});
+  assert(FM);
+
+  EXPECT_LT(*FM->getFeature("root"), *FM->getFeature("a"));
+  EXPECT_LT(*FM->getFeature("root"), *FM->getFeature("aa"));
+  EXPECT_LT(*FM->getFeature("root"), *FM->getFeature("b"));
+  EXPECT_LT(*FM->getFeature("a"), *FM->getFeature("aa"));
+  EXPECT_LT(*FM->getFeature("a"), *FM->getFeature("b"));
+  EXPECT_LT(*FM->getFeature("aa"), *FM->getFeature("b"));
```
vulder: The less than impl tries to implement a quite complicated relationship, could you please split this case up into different test cases that cover the different scenarios.

lib/Feature/FeatureModel.cpp
```diff
@@ -0,0 +1,148 @@
...
+}
+
+void FeatureModelBuilder::buildConstraints() {
+  for (const auto &F : Features.keys()) {
+    for (const auto &E : Excludes[F]) {
+      Features[F]->addExclude(Features[E].get());
+    }
+  }
+
+  for (const auto &C : Constraints) {
```
vulder: If we iterate over lists here we should use a better name than `C`. What is the actual type of `C`? maybe it could help to spell it out instead of auto.

lib/Feature/FeatureModel.cpp
```diff
@@ -0,0 +1,148 @@
...
+      if (B) {
+        for (const auto &P : C) {
+          for (const auto &PP : C) {
+            if (P.first != PP.first) {
+              Features[P.first]->addAlternative(Features[PP.first].get());
+            }
+          }
+        }
+      }
+    }
```
vulder, line 86: What happens in the else case? If `C.size()` is `0` or `1`?

lib/Feature/Feature.cpp
```diff
@@ -78,4 +52,45 @@ std::string NumericFeature::toString() const {
 
 std::string BinaryFeature::toString() const { return Feature::toString(); }
 
+bool Feature::operator<(const vara::feature::Feature &F) const {
```
vulder: For a `operator<` this is quite complex, so we should test this well, do the test cover all edge cases?

vulder (changes requested):
Found a few minor things

include/vara/Feature/FeatureModel.h
s9latimm: This method / builder is supposed to be only used while parsing / iteratively building a FM, so it makes not that much sense to build those connections for an already existing `Feature` by hand or even hold such an object in the first place. I added `addFeature(Feature&)` instead which basically behaves like inserting all related connections one after another, which can be used to duplicate existing subtrees between different FMs.

include/vara/Feature/Feature.h
s9latimm: FM consistency is assured through that Features will only be added with `FeatureModel::addFeature(std::unique_ptr<Feature>) : bool` checking whether a Feature can be inserted without violating any rules (currently only rudimentary). On the other hand it may be handy to create (unrelated) features beforehand to store information (not only for testing), building the actual FM afterwards (as the builder provides corresponding methods).

lib/Feature/Feature.cpp
s9latimm: I tested all pairwise relations on binary trees with height 2, basic cases and also unrelated features.

bindings/python/vara-feature/pybind_Feature.cpp
```diff
@@ -55,42 +47,37 @@ void init_feature_module_feature(py::module &M) {
...
-            return py::make_iterator(F.excludes_begin(), F.excludes_end());
+            return py::make_iterator(F.implications_begin(),
+                                     F.implications_end());
           },
           py::keep_alive<0, 1>(), R"pbdoc(Implicated features)pbdoc")
-      .def(
-          "add_implication", &vf::Feature::addImplication,
-          R"pbdoc(Implicate another Feature, i.e., selecting this feature implies the otherone.)pbdoc")
-      .def("implicates", &vf::Feature::implies,
+      .def("is_implied", &vf::Feature::isImplied,
```
vulder:
```suggestion
      .def("is_implied_by", &vf::Feature::isImplied,
```
This slight name change should make it a bit clearer

unittests/Feature/OrderedFeatureVector.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+TEST(OrderedFeatureVector, insertVariadic) {
+  auto FM = FeatureModelBuilder().buildSimpleFeatureModel(
+      {{"root", "a"}, {"root", "b"}});
+  OrderedFeatureVector OFV;
+
+  OFV.insert(FM->getFeature("a"), FM->getFeature("root"), FM->getFeature("b"));
+
+  EXPECT_EQ(OFV.size(), FM->size());
+  EXPECT_EQ(*OFV.begin(), FM->getFeature("root"));
```
vulder: All tests only check if the first one is root but do not verify if the others are in correct order.

include/vara/Feature/FeatureModel.h
```diff
@@ -1,96 +1,206 @@
...
+  /// Try to add \a NumericFeature.
+  bool addFeature(const std::string &Key,
+                  const NumericFeature::ValuesVariantType &Values,
+                  bool Opt = false,
+                  std::optional<FeatureSourceRange> Loc = std::nullopt) {
+    return Features
+        .try_emplace(Key, std::make_unique<NumericFeature>(Key, Values, Opt,
+                                                           std::move(Loc)))
+        .second;
+  }
```
vulder: We should merge these functions into a factory method and rename them to `make_feature` or `create_feature`, to not be confused with the `addFeature(Feature &)` method, which has a different behavior (Add a feature by copying it vs creating a new feature). Furthermore, `Key` is not a really good name here, it is the key for the map but from the API users side it is the feature name, and we should not confuse the user by naming things based on our implementation details.

Regarding the factory method:
```
    /// Try to create and add a new \a Feature to the \a FeatureModel.
  ///
  /// \param[in] FeatureName name of the \a Feature
  /// \param[in] FurtherArgs further arguments that should be passed to the
  ///                        \a Feature constructor
  ///
  /// \returns true, if the feature could be inserted into the \a FeatureModel
  template <
      typename FeatureTy, typename... Args,
      typename std::enable_if_t<std::is_base_of_v<Feature, FeatureTy>, int>>
  bool makeFeature(const std::string &FeatureName, Args... FurtherArgs) {
    return Features
        .try_emplace(FeatureName, std::make_unique<FeatureTy>(
                                      FeatureName, std::forward(FurtherArgs)...))
        .second;
  }
```
This allows users to create any kind of feature as long as it is derived from `Feature` and the passed args match one of the specific Features ctors.

include/vara/Feature/FeatureModel.h
```diff
@@ -1,96 +1,206 @@
...
+    Children[ParentKey].insert(Key);
+    Parents[Key] = ParentKey;
+    return this;
+  }
+
+  FeatureModelBuilder *addExclude(const std::string &Key,
+                                  const std::string &ExcludeKey) {
+    Excludes[Key].insert(ExcludeKey);
+    return this;
+  }
```
vulder: What happens in these functions if the `Key` is not a Feature name that was inserted before?

PS: `Key` is also a bad name for this param, we should rename this.

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -0,0 +1,51 @@
...
+class OrderedFeatureVector {
+public:
+  using ordered_feature_iterator = typename std::vector<Feature *>::iterator;
+  using const_ordered_feature_iterator =
+      typename std::vector<Feature *>::const_iterator;
+
+  OrderedFeatureVector() = default;
+
+  /// Insert feature while preserving ordering.
+  void insert(Feature *F);
```
vulder, line 23:
```suggestion
  void insert(Feature *F);
  
```

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -0,0 +1,51 @@
...
+  template <typename... Args> void insert(Feature *F, Args... FF) {
+    insert(F);
+    insert(FF...);
+  }
+  template <typename T> void insert(llvm::iterator_range<T> FF) {
```
vulder, line 30:
```suggestion

  template <typename T> void insert(llvm::iterator_range<T> FF) {
```

include/vara/Feature/Feature.h
```diff
@@ -21,44 +23,26 @@ namespace vara::feature {
...
-  FeatureListType Children;
-  FeatureListType Excludes;
-  FeatureListType Implications;
-  FeatureListType Alternatives;
-  RelationshipListTy Relationships;
-
-protected:
-  Feature(string Name, bool Opt, std::optional<Location> Loc)
-      : Name(std::move(Name)), Opt(Opt), Loc(std::move(Loc)) {}
+  enum FeatureKind { FK_BINARY, FK_NUMERIC };
```
vulder: Maybe it would be better to use a enum class here so integer constant can not be confused with the feature kinds.

vulder (changes requested):
Test coverage looks much better now :smiley:

include/vara/Feature/FeatureModel.h
s9latimm: Doing so is totally fine, as the consistency (existence) is asserted when building the final FM. As the parser can not assure that features are only referenced if they are already processed (like mutual exclusion), this can not be asserted in this method either.

padupr (approved):

vulder (approved):
lgtm

PR #15, opened 2020-06-16
Fix DTD

s9latimm:
While generating graphs for real VMs I encountered minor issues with missing / wrong tags, which will be added in this PR.


vulder (approved):
Ok, please put a note to our meeting doc that we should talk to Christian if we should implement this or not.

PR #17, opened 2020-07-30
Refactor FMFeature

s9latimm:
Replace `vara::Feature` with `vara::feature::Feature`.


include/vara/Feature/Feature.h
```diff
@@ -20,20 +25,27 @@ namespace vara::feature {
 //                               Feature Class
 //===----------------------------------------------------------------------===//
 
-/// \brief Base class for components of \a FeatureModel.
+/// \brief Base class for components of \a FeatureModel or \a FeatureRegion.
```
```suggestion
/// \brief Base class for components of \a FeatureModel.
```
We don't need to mention this here. The usage of VaRA should be "decoupled" from the feature lib.

vulder (approved):
lgtm

vulder (approved):

PR #18, opened 2020-09-02
Updated python bindings

padupr:
Updated python bindings and added xml writer

Closes se-passau/VaRa#667

### DIFF
--- a/bindings/python/vara-feature/pybind_FeatureModelXmlWriter.cpp
+++ b/bindings/python/vara-feature/pybind_FeatureModelXmlWriter.cpp
@@ -13,15 +13,15 @@ void init_xml_writer(py::module &M) {
   py::class_<vf::FeatureModelXmlWriter>(M, "FeatureModelXmlWriter")
       .def(py::init<const vf::FeatureModel &>())
       .def(
-          "dumps_feature_model",
+          "get_feature_model_as_string",
           [](vf::FeatureModelXmlWriter &Fmxw) {
             return Fmxw.writeFeatureModel();
           },
           R"pbdoc(Return the xml representation as string)pbdoc")
       .def(
-          "dump_feature_model",
+          "write_feature_model_to_file",
           [](vf::FeatureModelXmlWriter &Fmxw, const std::string StrS) {
             return Fmxw.writeFeatureModel(StrS);
           },
-          R"pbdoc(Write feature model to the given location)pbdoc");
+          R"pbdoc(Write feature model to the given file)pbdoc");
 }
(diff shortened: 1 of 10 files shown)


bindings/python/vara-feature/pybind_FeatureModelXmlWriter.cpp
```diff
@@ -0,0 +1,27 @@
+#include "pybind_common.h"
+#include "vara/Feature/FeatureModelWriter.h"
+
```
vulder:
```suggestion
#include "vara/Feature/FeatureModelWriter.h"
#include "pybind_common.h"

```

bindings/python/vara-feature/pybind_FeatureModelXmlWriter.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include <iostream>
+
+namespace vf = vara::feature;
+namespace py = pybind11;
+
+void init_xml_writer(py::module &M) {
+  py::class_<vf::FeatureModelXmlWriter>(M, "FeatureModelXmlWriter")
+      .def(py::init<const vf::FeatureModel &>())
+      .def(
+          "dumps_feature_model",
```
vulder: Why did you choose a different name here for the python side?

include/vara/Feature/Feature.h
```diff
@@ -159,6 +159,9 @@ class Feature {
   [[nodiscard]] FeatureSourceRange *getFeatureSourceRange() {
     return Source.has_value() ? &Source.value() : nullptr;
   }
+  void setFeatureSourceRange(FeatureSourceRange &Fsr) {
```
```suggestion
  void setFeatureSourceRange(FeatureSourceRange FeatureSR) {
```
If your going to move the object afterwards anyhow taking it as a value is prefered.

include/vara/Feature/Feature.h
```diff
@@ -159,6 +159,9 @@ class Feature {
...
+    Source = std::move(Fsr);
```
```suggestion
    Source = std::move(FeatureSR);
```

vulder (approved):
A few minor things but otherwise this looks good :)

bindings/python/vara-feature/pybind_FeatureModelXmlWriter.cpp
padupr: Python does not allow function overloading, so I used the naming of python's json lib to disambiguate the functions.

vulder: Yes, that's you use `dump_feature_model` vs `dumps_feature_model` but why did you change write to dump?
`dump_feature_model` could be `write_feature_model_to`
`dumps_feature_model` could be `get_feature_model_as_string` or `convert_feature_model_to_string`

PR #19, opened 2020-09-28
Location setter and comparison fixes

padupr:
Setting a location now copies the location and location equality now also requires an equal path.

### DIFF
--- a/bindings/python/tests/feature/test_feature.py
+++ b/bindings/python/tests/feature/test_feature.py
@@ -136,6 +136,36 @@ class TestFeature(unittest.TestCase):
         self.assertEqual(test_feature_0.location.end.column_offset,
                          18)
 
+        test_feature_0.location = test_feature_0.location
+        self.assertTrue(test_feature_0.location)
+        self.assertEqual(test_feature_0.location.path, path)
+        self.assertEqual(test_feature_0.location.start.line_number,
+                         4)
+        self.assertEqual(
+            test_feature_0.location.start.column_offset, 2)
+        self.assertEqual(test_feature_0.location.end.line_number, 4)
+        self.assertEqual(test_feature_0.location.end.column_offset,
+                         18)
+        
+    def test_feature_location_self_assign(self):
+        """ Checks if a feature's location is mutable."""
+        path = "test"
+        start_lco = feature.LineColumnOffset(4, 2)
+        end_lco = feature.LineColumnOffset(4, 18)
+        loc = feature.Location(path, start_lco, end_lco)
+        test_feature_0 = feature.BinaryFeature("Test", False, loc)
+
+        test_feature_0.location = test_feature_0.location
+        self.assertTrue(test_feature_0.location)
+        self.assertEqual(test_feature_0.location.path, path)
+        self.assertEqual(test_feature_0.location.start.line_number,
+                         4)
+        self.assertEqual(
+            test_feature_0.location.start.column_offset, 2)
+        self.assertEqual(test_feature_0.location.end.line_number, 4)
+        self.assertEqual(test_feature_0.location.end.column_offset,
+                         18)
+        
 
 class TestBinaryFeature(unittest.TestCase):
     """ Test BinaryFeature functionality.  """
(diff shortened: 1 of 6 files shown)


include/vara/Feature/Feature.h
```diff
@@ -159,8 +159,8 @@ class Feature {
   [[nodiscard]] FeatureSourceRange *getFeatureSourceRange() {
     return Source.has_value() ? &Source.value() : nullptr;
   }
-  void setFeatureSourceRange(FeatureSourceRange FeatureSR) {
-    Source = std::move(FeatureSR);
+  void setFeatureSourceRange(FeatureSourceRange &FeatureSR) {
+    Source = FeatureSourceRange(FeatureSR);
```
```suggestion
  void setFeatureSourceRange(FeatureSourceRange FeatureSR) {
    Source = std::move(FeatureSR);
```
The original should actually be correct. The FeatureSR is copied into this function, therefore, we are allowed to move it. If I revert you change here,. all feature-unitttests and python tests pass on my system.

Could you please revert it locally and double check if the test fails afterwards.

padupr: The original seems to work. I am not sure why it didn't work before. Maybe I forgot to update the pip package...

vulder: Maybe, but at least we now have a test for it :)

vulder (approved):

bindings/python/vara-feature/pybind_FeatureSourceRange.cpp
```diff
@@ -49,7 +49,12 @@ void init_feature_location_module(py::module &M) {
       .def_property_readonly(
           "end", &vf::FeatureSourceRange::getEnd,
           R"pbdoc(Get the end `LineColumnOffset` of this `Location`.)pbdoc")
-      .def("__str__", &vf::FeatureSourceRange::toString);
+      .def("__str__", &vf::FeatureSourceRange::toString)
+      .def("__eq__", [](const vf::FeatureSourceRange &Self,
+                        const vf::FeatureSourceRange *Other) {
+        if (!Other) return false;
```
vulder:
```suggestion
        if (!Other) {
            return false;
        }
```
please always use brackets, a if without is a bug ready to happen

bindings/python/vara-feature/pybind_FeatureSourceRange.cpp
```diff
@@ -63,9 +68,9 @@ void init_feature_location_module(py::module &M) {
...
           &vf::FeatureSourceRange::FeatureSourceLocation::setColumnOffset)
       .def("__str__", &vf::FeatureSourceRange::FeatureSourceLocation::toString)
-      .def("__eq__",
-           [](const vf::FeatureSourceRange::FeatureSourceLocation &Self,
-              const vf::FeatureSourceRange::FeatureSourceLocation &Other) {
-             return Self == Other;
-           });
+      .def("__eq__", [](const vf::FeatureSourceRange::FeatureSourceLocation &Self,
+                        const vf::FeatureSourceRange::FeatureSourceLocation *Other) {
+        if (!Other) return false;
```
vulder:
```suggestion
        if (!Other) {
            return false
        };
```
same

bindings/python/tests/feature/test_feature.py
```diff
@@ -136,6 +136,36 @@ def test_feature_location_setters(self):
...
+        test_feature_0.location = test_feature_0.location
+        self.assertTrue(test_feature_0.location)
+        self.assertEqual(test_feature_0.location.path, path)
+        self.assertEqual(test_feature_0.location.start.line_number,
+                         4)
+        self.assertEqual(
+            test_feature_0.location.start.column_offset, 2)
+        self.assertEqual(test_feature_0.location.end.line_number, 4)
+        self.assertEqual(test_feature_0.location.end.column_offset,
+                         18)
```
vulder: This test code looks like the test case bellow. Does this make sense here?
If it does, this should be a different test case which specifically checks for self assignment for location setters.

vulder (changes requested):

vulder (approved):

PR #22, opened 2020-10-06
added explicit casts from stringref to string

Kallistos:
Since LLVM-11 casts from stringref to string have to be explicit

added explicit cast from stringref to string in /lib/Feature/FeatureModel.cpp and /tools/fm-viewer/FeatureModelViewer.cpp

### DIFF
--- a/lib/Feature/FeatureModel.cpp
+++ b/lib/Feature/FeatureModel.cpp
@@ -12,7 +12,7 @@ void FeatureModel::dump() const {
 
 bool FeatureModel::addFeature(std::unique_ptr<Feature> Feature) {
   // TODO(s9latimm): check consistency
-  std::string FeatureName = Feature->getName();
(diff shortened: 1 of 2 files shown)


lib/Feature/FeatureModel.cpp
```diff
@@ -12,7 +12,7 @@ void FeatureModel::dump() const {
 
 bool FeatureModel::addFeature(std::unique_ptr<Feature> Feature) {
   // TODO(s9latimm): check consistency
-  std::string FeatureName = Feature->getName();
+  std::string FeatureName = std::string(Feature->getName());
```
vulder:
```suggestion
  auto FeatureName = std::string(Feature->getName());
```

vulder (approved):
looks good

just a small nit, in case we spell the type on the rhs we don't need it on the lhs anymore

PR #24, opened 2020-10-21
Refactor constraints and add feature groups

s9latimm:


include/vara/Feature/Constraint.h
```diff
@@ -0,0 +1,64 @@
...
+  Constraint(ConstraintKind Kind) : Kind(Kind) {}
+  Constraint(const Constraint &) = delete;
+  Constraint &operator=(const Constraint &) = delete;
+  virtual ~Constraint() = default;
+
+  [[nodiscard]] ConstraintKind getKind() const { return Kind; };
+
+private:
+  friend class FeatureModel;
+  friend class FeatureModelBuilder;
```
vulder: Why do constraints need private member access to a FeatureModel or the Builder?

include/vara/Feature/Constraint.h
```diff
@@ -0,0 +1,64 @@
...
+};
+
+class UnaryConstraint : public Constraint {
+public:
+  UnaryConstraint(string Operator, Constraint *Operand)
+      : Operator(std::move(Operator)), Operand(Operand),
+        Constraint(ConstraintKind::CK_UNARY) {}
+
+private:
+  string Operator;
```
vulder: Why are Operators strings?

include/vara/Feature/Constraint.h
```diff
@@ -0,0 +1,64 @@
...
+
+class BinaryConstraint : public Constraint {
+public:
+  BinaryConstraint(string Operator, Constraint *LeftOperand,
+                   Constraint *RightOperand)
+      : Operator(std::move(Operator)), LeftOperand(LeftOperand),
+        RightOperand(RightOperand), Constraint(ConstraintKind::CK_BINARY) {}
+
+private:
+  string Operator;
```
vulder: See, below

include/vara/Feature/Constraint.h
```diff
@@ -0,0 +1,64 @@
...
+#include "vara/Feature/Feature.h"
+
+namespace vara::feature {
+//===----------------------------------------------------------------------===//
+//                               Constraint
+//===----------------------------------------------------------------------===//
+
+/// \brief Tree like representation of constraints between features
+
+class Constraint {
```
vulder, line 23: I dont fully understand this constraint design. There is no public interface that users can use so what is the planed purpose of the Constraint class?

include/vara/Feature/Feature.h
```diff
@@ -51,109 +48,15 @@ class Feature {
...
-
-  //===--------------------------------------------------------------------===//
-  // Default
-
-  feature_iterator begin() { return children_begin(); }
-  feature_iterator end() { return children_end(); }
-  [[nodiscard]] const_feature_iterator begin() const {
-    return children_begin();
+  [[nodiscard]] Feature *getParentFeature() const {
+    for (auto *P = this->getParent(); P; P = P->getParent()) {
```
```suggestion
    for (FeatureTreeNode *P = this->getParent(); P; P = P->getParent()) {
```
It's not quite clear here with auto. One could expect here that getParent returns a `Feature`.

include/vara/Feature/Feature.h
```diff
@@ -51,109 +48,15 @@ class Feature {
...
+      auto *F = llvm::dyn_cast<Feature>(P);
+      if (F) {
+        return F;
+      }
```
vulder, line 60:
```suggestion
      if (llvm::isa<Feature>(P)) {
        return P;
      }
```

include/vara/Feature/Feature.h
```diff
@@ -51,109 +48,15 @@ class Feature {
...
+    }
+    return nullptr;
```
vulder, line 62:
```suggestion
    assert(false && "Node had no parent but was not the root node.");
    return nullptr;
```
related to above

include/vara/Feature/Feature.h
```diff
@@ -51,109 +48,15 @@ class Feature {
...
-  feature_iterator end() { return children_end(); }
-  [[nodiscard]] const_feature_iterator begin() const {
-    return children_begin();
+  [[nodiscard]] Feature *getParentFeature() const {
+    for (auto *P = this->getParent(); P; P = P->getParent()) {
+      auto *F = llvm::dyn_cast<Feature>(P);
+      if (F) {
+        return F;
+      }
+    }
```
vulder, line 61:
```suggestion
      if (P->isRoot()) {
        return nullptr;
      }
    }
```
We could add an additional safty check here, not sure if it is worth it. What do you think?

include/vara/Feature/FeatureModel.h
```diff
@@ -21,7 +23,7 @@ class FeatureModel {
 public:
   using FeatureMapTy = llvm::StringMap<std::unique_ptr<Feature>>;
   using OrderedFeatureTy = OrderedFeatureVector;
-  using ConstraintTy = llvm::SmallVector<std::pair<std::string, bool>, 3>;
+  using ConstraintTy = Constraint;
   using ConstraintsTy = std::vector<ConstraintTy>;
```
```suggestion
  using ConstraintContainerTy = std::vector<ConstraintTy>;
```
I think this could be a better name for what we do with the type.

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -0,0 +1,88 @@
...
+
+//===----------------------------------------------------------------------===//
+//                          FeatureTreeNode Class
+//===----------------------------------------------------------------------===//
+
+class FeatureTreeNode : private llvm::DGNode<FeatureTreeNode, FeatureTreeNode> {
+public:
+  enum class NodeKind { NK_FEATURE, NK_GROUP };
+
+  using NodeSetType = typename std::set<FeatureTreeNode *>;
```
vulder: Hmm, if we expect only a few child nodes we could use a SmallSet here with an appropriate size.

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -0,0 +1,88 @@
...
+  //===--------------------------------------------------------------------===//
+  // Children
+
+  [[nodiscard]] iterator begin() { return DGNode::begin(); }
+  [[nodiscard]] const_iterator begin() const { return DGNode::begin(); }
+
+  [[nodiscard]] iterator end() { return DGNode::end(); }
+  [[nodiscard]] const_iterator end() const { return DGNode::end(); }
+
+  llvm::iterator_range<iterator> children() {
```
vulder:
```suggestion
  [[nodiscard]] llvm::iterator_range<iterator> children() {
```

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -32,8 +32,11 @@ class OrderedFeatureVector {
   }
 
   template <typename T> void insert(llvm::iterator_range<T> FF) {
-    for (const auto &F : FF) {
-      insert(F);
+    for (const auto &N : FF) {
```
vulder: `FF` and you get a `N` when iterating over it. We could use better names here :smile:

include/vara/Feature/Relationship.h
```diff
@@ -0,0 +1,46 @@
...
+//===----------------------------------------------------------------------===//
+
+class Relationship : public FeatureTreeNode {
+public:
+  enum class RelationshipKind { RK_ALTERNATIVE, RK_OR };
+
+  Relationship(RelationshipKind Kind)
+      : Kind(Kind), FeatureTreeNode(NodeKind::NK_GROUP) {}
+
+  [[nodiscard]] RelationshipKind getRelationshipKind() const { return Kind; }
```
vulder: I, normaly, use just `getKind` for kind accessers. I think this should be the llvm way, but we should double check to be sure.

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -0,0 +1,88 @@
...
+  enum class NodeKind { NK_FEATURE, NK_GROUP };
+
+  using NodeSetType = typename std::set<FeatureTreeNode *>;
+
+  FeatureTreeNode(NodeKind Kind) : Kind(Kind){};
+  FeatureTreeNode(const FeatureTreeNode &) = delete;
+  FeatureTreeNode &operator=(const FeatureTreeNode &) = delete;
+  virtual ~FeatureTreeNode() = default;
+
+  [[nodiscard]] NodeKind getNodeKind() const { return Kind; }
```
vulder: Same as below `getKind`

lib/Feature/Feature.cpp
```diff
@@ -11,24 +11,15 @@ std::string Feature::toString() const {
...
-  StrS << "],\n  implies: [";
-  for (const auto &Implication : Implications) {
-    StrS << Implication->Name << ",";
-  }
-  StrS << "],\n  alternatives: [";
-  for (const auto &Alternative : Alternatives) {
-    StrS << Alternative->Name << ",";
+  for (const auto &Child : children()) {
+    auto *F = llvm::dyn_cast<Feature>(Child);
+    if (F) {
```
vulder: use `llvm::isa`

lib/Feature/FeatureModel.cpp
```diff
@@ -18,73 +18,74 @@ bool FeatureModel::addFeature(std::unique_ptr<Feature> Feature) {
...
-    Child->setParent(Value);
+    Child->setParent(*Value);
   }
   if (Value->isRoot()) {
-    if (*Root->getParent() == *Value) {
+    if (*Root->getParentFeature() == *Value) {
       Root = Value;
     } else {
-      Value->setParent(Root);
+      Value->setParent(*Root);
```
vulder: If we have a lot of these uses we could add an setParent overload for pointers.

lib/Feature/FeatureModel.cpp
```diff
@@ -167,35 +169,44 @@ bool FeatureModelBuilder::addFeature(Feature &F) {
...
+                                    Loc)) {
       return false;
     }
     break;
   case Feature::FeatureKind::FK_NUMERIC:
     if (!makeFeature<NumericFeature>(
-            std::string(F.getName()), dynamic_cast<NumericFeature *>(&F)->getValues(),
-            F.isOptional(), Loc)) {
+            std::string(F.getName()),
+            dynamic_cast<NumericFeature *>(&F)->getValues(), F.isOptional(),
```
vulder: Hmm, I'm wondering aren't we using the llvm kind/classof infra for Feature? So, this shoudl/could be a `llvm::dyn_cast`?

lib/Feature/FeatureModel.cpp
```diff
@@ -107,8 +108,8 @@ bool FeatureModelBuilder::buildTree(const string &FeatureName,
     if (!buildTree(Child, Visited)) {
       return false;
     }
-    Features[FeatureName]->addChild(Features[Child].get());
-    Features[Child]->setParent(Features[FeatureName].get());
+    Features[FeatureName]->addEdge(*Features[Child]);
```
vulder: Same here

vulder (changes requested):
I did a first pass over the current changes, looks good overall, I like the llvm DGNode choice.

include/vara/Feature/Feature.h
s9latimm: Current implementation of `isRoot` just checks `Parent == nullptr`, so one would already expect, that `getParent` on a root node returns a  `nullptr` (which is handled in the loop condition).

include/vara/Feature/Feature.h
s9latimm: If a node has no parent `isRoot` always returns true. See my comment regarding checking this inside the loop. Currently we guarantee consistency only when inside a FMTree (with a single root).

lib/Feature/Feature.cpp
s9latimm: I call `F->getName()` afterwards, which needs this `Feature`.

lib/Feature/FeatureModel.cpp
s9latimm: It should have been. :laughing:

include/vara/Feature/Feature.h
s9latimm: This method should return an actual parent `Feature` (if exists), but `P` is only a `FeatureTreeNode` at this point.

lib/Feature/Feature.cpp
vulder: A yes, I missed that.

include/vara/Feature/Constraint.h
```diff
@@ -25,55 +25,166 @@ class Constraint {
...
-    case OperatorKind::OK_AND: return "&";
-    case OperatorKind::OK_OR: return "v";
-    case OperatorKind::OK_EQUALS: return "=";
-    case OperatorKind::OK_IMPLIES: return "=>";
-    case OperatorKind::OK_EQUIVALENCE: return "<=>";
-    case OperatorKind::OK_XOR: return "^";
-    default: return "";
-    }
-  };
+class BoolschesConstraint {
```
```suggestion
class BooleanConstraint {
```
This should be spell in English I guess :)

include/vara/Feature/Constraint.h
```diff
@@ -25,55 +25,166 @@ class Constraint {
...
 class BinaryConstraint : public Constraint {
 public:
-  BinaryConstraint(ConstraintOperator Operator, Constraint *LeftOperand,
-                   Constraint *RightOperand)
-      : Operator(Operator), LeftOperand(LeftOperand),
+  BinaryConstraint(Constraint *LeftOperand, Constraint *RightOperand)
+      : LeftOperand(LeftOperand),
         RightOperand(RightOperand), Constraint(ConstraintKind::CK_BINARY) {}
 
+  [[nodiscard]] virtual std::string toString() const;
```
```suggestion
  [[nodiscard]] std::string toString() const override;
```

include/vara/Feature/Constraint.h
```diff
@@ -25,55 +25,166 @@ class Constraint {
...
 };
 
 class UnaryConstraint : public Constraint {
 public:
-  UnaryConstraint(ConstraintOperator Operator, Constraint *Operand)
-      : Operator(Operator), Operand(Operand),
+  UnaryConstraint(Constraint *Operand) : Operand(Operand),
         Constraint(ConstraintKind::CK_UNARY) {}
 
+  [[nodiscard]] virtual std::string toString() const;
```
```suggestion
  [[nodiscard]] std::string toString() const override;
```

include/vara/Feature/Constraint.h
```diff
@@ -25,55 +25,166 @@ class Constraint {
   ConstraintKind Kind;
 };
```
vulder, line 80: Github does not let me annotate the lines above :cry: 
Both sub hierachies have the `toString` method so should be not put it into the base class?

Add this to the public API
```cpp
[[nodiscard]] virtual std::string toString() const;
```

include/vara/Feature/Constraint.h
```diff
@@ -25,55 +25,166 @@ class Constraint {
...
+
+class OrConstraint : public BinaryConstraint, public BoolschesConstraint {
+public:
+  OrConstraint(Constraint *LeftOperand, Constraint *RightOperand)
+      : BinaryConstraint(LeftOperand, RightOperand) {}
+
+  [[nodiscard]] std::string toString() const override;
+};
+
+class XOrConstraint : public BinaryConstraint, public BoolschesConstraint {
```
```suggestion
class XorConstraint : public BinaryConstraint, public BoolschesConstraint {
```
Maybe spellit `Xor` or `XOR`, spelling it `XOr` looks very strange to me :laughing:

include/vara/Feature/Constraint.h
```diff
@@ -25,55 +25,166 @@ class Constraint {
...
+
+class GreaterEquelsConstraint : public BinaryConstraint, public NumericConstraint {
+public:
+  GreaterEquelsConstraint(Constraint *LeftOperand, Constraint *RightOperand)
+      : BinaryConstraint(LeftOperand, RightOperand) {}
+
+  [[nodiscard]] std::string toString() const override;
+};
+
 class PrimaryConstraint : public Constraint {
```
vulder: What is the reasoning behind `Primary` as a constraint name? I'm not familiar with this.

include/vara/Feature/Constraint.h
Kallistos: If i change this to 
```suggestion
  [[nodiscard]] virtual std::string toString() const;
```
there is no virtual function, which means the class is not abstract, right?

include/vara/Feature/Constraint.h
Kallistos: same as above

include/vara/Feature/Constraint.h
Kallistos: you're right! but how can i mark both subclasses as abstract? (BooleanConstraint and NumericConstraint?)

.travis.yml
```diff
@@ -12,43 +12,43 @@ jobs:
...
-            - clang-tidy-10
-      env: CXX=clang++-10 BUILD_TYPE=Debug
+            - clang-tidy-11
+      env: CXX=clang++-10 BUILD_TYPE=Debug LLVM_VERSION=10
     - name: Clang-10 Release
-      stage: Latest
+      stage: Legacy
       addons: *clang10
-      env: CXX=clang++-10 BUILD_TYPE=Release
+      env: CXX=clang++-9 BUILD_TYPE=Release LLVM_VERSION=10
```
vulder:
```suggestion
      env: CXX=clang++-10 BUILD_TYPE=Release LLVM_VERSION=10
```

include/vara/Feature/Constraint.h
Kallistos: Primary represents the leave nodes in the constraint tree, i based this on the expression hierarchy from C; there is a primary expression either a constant, a literal and so on. In our case this is right now either a number or  a feature.

include/vara/Feature/FeatureModel.h
```diff
@@ -96,8 +97,11 @@ class FeatureModel {
   string Name;
   fs::path Path;
   FeatureMapTy Features;
+  ConstraintContainerTy Constraints;
   Feature *Root;
 
+  FeatureModel() = default;
```
vulder, line 107: Why do we add a default ctor for FeatureModels? Before we always wanted to have at least a path and a name.

include/vara/Feature/Constraint.h
```diff
@@ -37,161 +45,216 @@ class NumericConstraint {
...
+
+  std::variant<Feature *, std::string> getFeature() { return FV; }
+
+  void accept(ConstraintVisitor &V);
 
 private:
-  Feature *Feature;
+  friend FeatureModelBuilder;
+
+  std::variant<Feature *, std::string> FV;
```
vulder: What is the reasoning behind the string alternative?

include/vara/Feature/Constraint.h
```diff
@@ -37,161 +45,216 @@ class NumericConstraint {
...
+class FeatureModelBuilder;
+
 class PrimaryConstraint : public Constraint {
 public:
-  PrimaryConstraint(Feature *Feature)
-      : Feature(Feature), Constraint(ConstraintKind::CK_PRIMARY) {}
+  PrimaryConstraint(std::variant<Feature *, std::string> FV)
+      : FV(std::move(FV)), Constraint(ConstraintKind::CK_PRIMARY) {}
+
+  std::variant<Feature *, std::string> getFeature() { return FV; }
```
vulder: This interface seems a bit off. One requests a Feature but also could get a string (currently I'm not sure what the meaning of the string is). Depending on the string meaning, it could be better to check the type in the variant here and return an optional feature (or nullptr) if there was a string in the variant and the `Feature *` otherwise. We could then offer the same interface for the "meaning of string".

This would not tie the user into our variant usage and require all using code to make variant dispatches.

include/vara/Feature/Constraint.h
vulder: As long as they don't offer a `toString` method them selfs (or higher up the hierarchy) they are still abstract.

include/vara/Feature/FeatureModel.h
s9latimm: This way `FeatureModelBuilder` can inherit the data structure without already setting publicly required fields (like path / name) as they are build afterwards.

include/vara/Feature/Constraint.h
s9latimm: While parsing  / building the constraint tree, the actual feature might not be present in the FM, so the name is used as placeholder and replaced afterwards. On the other hand we are able to build a constraint tree without requiring any feature parsing.

include/vara/Feature/Constraint.h
s9latimm: Indeed, it seems reasonable to wrap the optional feature within the getter.

.travis.yml
```diff
@@ -13,42 +13,42 @@ jobs:
...
           packages:
-            - llvm-10-dev
-            - clang++-10
-            - clang-tidy-10
-      env: CXX=clang++-10 BUILD_TYPE=Debug
-    - name: Clang-10 Release
+            - llvm-11-dev
+            - clang++-11
+            - clang-tidy-11
+      env: CXX=clang++-11 BUILD_TYPE=Debug LLVM_VERSION=11
```
vulder: Is the `LLVM_VERSION=11` needed? Normally, it should default to the current one, meaning be automatically 11.

bindings/python/vara-feature/pybind_Feature.cpp
```diff
@@ -24,62 +24,67 @@ void init_feature_module_feature(py::module &M) {
...
+      //          "is_alternative", &vf::Feature::isAlternative,
+      //          R"pbdoc(Checks if a Feature is an alternative of this
+      //          Feature.)pbdoc")
+      //
+      //      .def_property("location", &vf::Feature::getFeatureSourceRange,
+      //                    &vf::Feature::setFeatureSourceRange,
+      //                    py::return_value_policy::reference,
+      //                    R"pbdoc(Returns the code location of the feature
+      //                    variable if
+      //                possible, otherwise, `None`.)pbdoc")
```
vulder: Do we have new iterators for these? Or should this be removed?

include/vara/Feature/Constraint.h
```diff
@@ -0,0 +1,379 @@
...
+class FeatureModelBuilder;
+
+class PrimaryConstraint : public Constraint {
+public:
+  PrimaryConstraint() : Constraint(ConstraintKind::CK_PRIMARY) {}
+};
+
+class PrimaryIntegerConstraint : public PrimaryConstraint {
+public:
+  PrimaryIntegerConstraint(int V) : V(V) {}
```
vulder: needs member accessor

vulder: probably needs kind

include/vara/Feature/Constraint.h
```diff
@@ -0,0 +1,379 @@
...
+//===----------------------------------------------------------------------===//
+//                               Constraint
+//===----------------------------------------------------------------------===//
+
+class ConstraintVisitor;
+
+/// \brief Tree like representation of constraints between features
+class Constraint {
+public:
+  enum class ConstraintKind { CK_BINARY, CK_UNARY, CK_PRIMARY };
```
vulder: create specific kinds for sub types

lib/Feature/FeatureModelParser.cpp
```diff
@@ -33,7 +33,28 @@ bool FeatureModelXmlParser::parseConfigurationOption(xmlNode *Node,
             if (!xmlStrcmp(Child->name, XmlConstants::OPTIONS)) {
               std::unique_ptr<xmlChar, void (*)(void *)> CCnt(
                   xmlNodeGetContent(Child), xmlFree);
-              FMB.addExclude(Name, reinterpret_cast<char *>(CCnt.get()));
+              FMB.addConstraint(std::make_unique<ImpliesConstraint>(
```
vulder: Use extra `ExcludeConstraint`

lib/Feature/FeatureModelParser.cpp
```diff
@@ -106,17 +129,18 @@ bool FeatureModelXmlParser::parseConstraints(xmlNode *Node) {
                                          .get()));
         const std::regex Regex(R"((!?\w+))");
```
vulder: Use llvm::Regex

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -0,0 +1,91 @@
...
+    return llvm::make_range(begin(), end());
+  }
+
+  [[nodiscard]] bool hasEdgeTo(const FeatureTreeNode &N) const {
+    return std::find(begin(), end(), &N) != end();
+  }
+
+protected:
+  FeatureTreeNode(NodeKind Kind, FeatureTreeNode *Parent,
+                  const NodeSetType &Children)
```
vulder:
```suggestion
                  const llvm::SmallSetImpl<FeatureTreeNode *> &Children)
```

include/vara/Feature/Constraint.h
```diff
@@ -0,0 +1,379 @@
...
+  }
+
+  [[nodiscard]] virtual std::string toString() const = 0;
+
+  [[nodiscard]] virtual std::string toHTML() const { return toString(); }
+
+  virtual void accept(ConstraintVisitor &V) = 0;
+
+private:
+  ConstraintKind Kind;
```
```suggestion
  const ConstraintKind Kind;
```
So we can guarantee that the kind does not change during the Lifetime of a constraint.

include/vara/Feature/Constraint.h
```diff
@@ -0,0 +1,379 @@
...
+class ConstraintVisitor;
+
+/// \brief Tree like representation of constraints between features
+class Constraint {
+public:
+  enum class ConstraintKind { CK_BINARY, CK_UNARY, CK_PRIMARY };
+
+  Constraint(ConstraintKind Kind) : Kind(Kind) {}
+  Constraint(const Constraint &) = delete;
+  Constraint &operator=(const Constraint &) = delete;
```
```suggestion
  Constraint &operator=(const Constraint &) = delete;
  Constraint(Constraint &&) = delete;
  Constraint &operator=(Constraint &&) = delete;
```
If we don't implement massign and mctor our selfs they will be disabled, so we should at least flag them as deleted here, or implement them.

include/vara/Feature/Feature.h
```diff
@@ -188,39 +107,28 @@ class Feature {
...
-
-  void setParent(Feature *Feature) { Parent = Feature; }
-
-  void addExclude(Feature *F) { Excludes.insert(F); }
-
-  void addAlternative(Feature *F) { Alternatives.insert(F); }
-
-  void addImplication(Feature *F) { Implications.insert(F); }
-
   FeatureKind Kind;
```
```suggestion
  const FeatureKind Kind;
```

include/vara/Feature/Feature.h
```diff
@@ -188,39 +107,28 @@ class Feature {
...
   string Name;
   bool Opt;
```
vulder: I would move the `bool` flag to the end of the member list.

include/vara/Feature/Feature.h
```diff
@@ -21,25 +22,27 @@ using std::string;
...
 
   enum class FeatureKind { FK_BINARY, FK_NUMERIC, FK_UNKNOWN };
 
   Feature(std::string Name)
-      : Kind(FeatureKind::FK_UNKNOWN), Name(std::move(Name)), Opt(false),
-        Source(std::nullopt), Parent(nullptr) {}
+      : FeatureTreeNode(NodeKind::NK_FEATURE), Kind(FeatureKind::FK_UNKNOWN),
+        Name(std::move(Name)), Opt(false), Source(std::nullopt) {}
   Feature(const Feature &) = delete;
   Feature &operator=(const Feature &) = delete;
```
vulder, line 38: Same with the mctor/massign here.

include/vara/Feature/Feature.h
```diff
@@ -180,6 +91,14 @@ class Feature {
     Source = std::move(FeatureSR);
   }
 
+  llvm::iterator_range<constraint_iterator> constraints() {
```
```suggestion
  [[nodiscard]] llvm::iterator_range<constraint_iterator> constraints() {
```

include/vara/Feature/FeatureModel.h
```diff
@@ -357,9 +323,14 @@ template <> struct GraphWriter<vara::feature::FeatureModel *> {
       for (auto *Child : *Node) {
         emitCluster(Child, Indent + 2);
         O.indent(Indent + 2);
-        emitEdge(Node, Child,
-                 llvm::formatv("arrowhead={0}",
-                               Child->isOptional() ? "odot" : "dot"));
+        auto *F = llvm::dyn_cast<vara::feature::Feature>(Child);
+        if (F) {
```
```suggestion
        
        if (auto *F = llvm::dyn_cast<vara::feature::Feature>(Child); F) {
```
If you only use the value inside the if, you can use a if initializer. Makes for better scoping. `F` will be out of scope after the `if`, making it impossible to be used after wards, which would be good in the case `F == nullptr`

include/vara/Feature/FeatureModel.h
```diff
@@ -371,19 +342,47 @@ template <> struct GraphWriter<vara::feature::FeatureModel *> {
...
-        "table>>",
-        DOT::EscapeString(Node->getName().str()),
-        (Node->getFeatureSourceRange()
-             ? "</td></tr><hr/><tr><td>" +
-                   DOT::EscapeString(Node->getFeatureSourceRange()->toString())
-             : ""));
-
+    std::string Label;
+    auto *F = llvm::dyn_cast<vara::feature::Feature>(Node);
+    if (F) {
```
```suggestion
    if (auto *F = llvm::dyn_cast<vara::feature::Feature>(Node); F) {
```

include/vara/Feature/FeatureModel.h
```diff
@@ -371,19 +342,47 @@ template <> struct GraphWriter<vara::feature::FeatureModel *> {
...
+           << "</td></tr>";
+      }
+      Label =
+          llvm::formatv(
+              "<<table align=\"center\" valign=\"middle\" border=\"0\" "
+              "cellborder=\"0\" "
+              "cellpadding=\"5\">{0}{1}{2}</table>>",
+              llvm::formatv("<tr><td><b>{0}</b></td></tr>",
+                            DOT::EscapeString(F->getName().str()))
+                  .str(),
```
```suggestion
                  ,
```

include/vara/Feature/FeatureModel.h
```diff
@@ -371,19 +342,47 @@ template <> struct GraphWriter<vara::feature::FeatureModel *> {
...
+              CS.str(),
+              (F->getFeatureSourceRange()
+                   ? llvm::formatv("<hr/><tr><td>{0}</td></tr>",
+                                   DOT::EscapeString(
+                                       F->getFeatureSourceRange()->toString()))
+                         .str()
+                   : ""))
+              .str();
```
```suggestion
              ;
```
I think we don't need to convert these objects to strings before using them. I'm not 100% sure so please double check :)

include/vara/Feature/FeatureModel.h
```diff
@@ -371,19 +342,47 @@ template <> struct GraphWriter<vara::feature::FeatureModel *> {
...
+    } else {
+      auto *R = llvm::dyn_cast<vara::feature::Relationship>(Node);
+      if (R) {
```
```suggestion
      if (auto *R = llvm::dyn_cast<vara::feature::Relationship>(Node); R) {
```

lib/Feature/Feature.cpp
vulder: Some here

lib/Feature/FeatureModel.cpp
```diff
@@ -166,35 +154,36 @@ bool FeatureModelBuilder::addFeature(Feature &F) {
...
-  for (const auto *Exclude : F.excludes()) {
-    this->addExclude(std::string(F.getName()), std::string(Exclude->getName()));
-  }
-  for (const auto *Alternative : F.alternatives()) {
-    this->addConstraint({{std::string(F.getName()), true}, {std::string(Alternative->getName()), true}});
-  }
-  for (const auto *Implication : F.implications()) {
-    this->addConstraint({{std::string(F.getName()), false}, {std::string(Implication->getName()), true}});
+    const auto *C = llvm::dyn_cast<vara::feature::Feature>(Child);
+    if (C) {
```
vulder: same here

lib/Feature/FeatureModelParser.cpp
```diff
@@ -33,7 +33,28 @@ bool FeatureModelXmlParser::parseConfigurationOption(xmlNode *Node,
             if (!xmlStrcmp(Child->name, XmlConstants::OPTIONS)) {
               std::unique_ptr<xmlChar, void (*)(void *)> CCnt(
                   xmlNodeGetContent(Child), xmlFree);
-              FMB.addExclude(Name, reinterpret_cast<char *>(CCnt.get()));
+              FMB.addConstraint(std::make_unique<ImpliesConstraint>(
+                  std::make_unique<PrimaryFeatureConstraint>(
+                      std::make_unique<Feature>(Name)),
+                  std::make_unique<NotConstraint>(
```
```suggestion
                  make_unique<NotConstraint>(
```
Add a using decl at the beginning of the file.
```
using std::make_unique;
```
This should make the code shorter and a bit easier to read.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -96,7 +117,9 @@ bool FeatureModelXmlParser::parseOptions(xmlNode *Node, bool Num = false) {
   return true;
 }
 
-bool FeatureModelXmlParser::parseConstraints(xmlNode *Node) {
+// TODO(s9latimm): Refactor with new Constraints representation
```
vulder: If you want to keep the TODO for future reference, please relate it to the issue that implements the constraint parser.

```
// TODO(se-passau/VaRA#666): Comment
```

lib/Feature/FeatureModelWriter.cpp
```diff
@@ -140,7 +140,10 @@ int FeatureModelXmlWriter::writeNumericFeatures(xmlTextWriterPtr Writer) {
   return RC;
 }
 
-int FeatureModelXmlWriter::writeBooleanConstraints(xmlTextWriterPtr Writer) {
+// TODO(s9latimm): Refactor with new Constraints representation
```
vulder: same

lib/Feature/FeatureModelWriter.cpp
```diff
@@ -223,39 +225,41 @@ int FeatureModelXmlWriter::writeFeature(xmlTextWriterPtr Writer,
     CHECK_RC
   }
 
+  // TODO(s9latimm): Refactor with new Constraints representation
```
vulder: same

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -0,0 +1,91 @@
...
+  friend class FeatureModel;
+  friend class FeatureModelBuilder;
+
+  void setParent(FeatureTreeNode &Feature) { Parent = &Feature; }
+  void setParent(FeatureTreeNode *Feature) { Parent = Feature; }
+
+  bool addEdge(FeatureTreeNode *Feature) { return DGNode::addEdge(*Feature); }
+
+  FeatureTreeNode *Parent{nullptr};
+  NodeKind Kind;
```
vulder:
```suggestion
  const NodeKind Kind;
```

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -0,0 +1,91 @@
...
+  [[nodiscard]] NodeKind getKind() const { return Kind; }
+
+  //===--------------------------------------------------------------------===//
+  // Parent
+
+  [[nodiscard]] FeatureTreeNode *getParent() const { return Parent; }
+
+  [[nodiscard]] bool isRoot() const { return Parent == nullptr; }
+
+  bool hasEdgeFrom(FeatureTreeNode &N) const { return Parent == &N; }
```
vulder:
```suggestion
  [[nodiscard]] bool hasEdgeFrom(FeatureTreeNode &N) const { return Parent == &N; }
```

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -0,0 +1,91 @@
...
+
+class FeatureTreeNode : private llvm::DGNode<FeatureTreeNode, FeatureTreeNode> {
+public:
+  enum class NodeKind { NK_FEATURE, NK_RELATIONSHIP };
+
+  using NodeSetType = typename llvm::SmallSet<FeatureTreeNode *, 3>;
+
+  FeatureTreeNode(NodeKind Kind) : Kind(Kind){};
+  FeatureTreeNode(const FeatureTreeNode &) = delete;
+  FeatureTreeNode &operator=(const FeatureTreeNode &) = delete;
```
vulder: mctor/massign

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -0,0 +1,91 @@
...
+
+#include "llvm/ADT/DirectedGraph.h"
+#include "llvm/ADT/SmallSet.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/ADT/iterator_range.h"
+#include "llvm/IR/Function.h"
+#include "llvm/IR/Value.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <llvm/ADT/SetVector.h>
```
vulder:
```suggestion
#include "llvm/ADT/SetVector.h"

```
+ include sorting

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -31,9 +31,12 @@ class OrderedFeatureVector {
     insert(FF...);
   }
 
-  template <typename T> void insert(llvm::iterator_range<T> FF) {
-    for (const auto &F : FF) {
-      insert(F);
+  template <typename T> void insert(llvm::iterator_range<T> Iter) {
+    for (const auto &Ptr : Iter) {
+      auto *F = llvm::dyn_cast<Feature>(Ptr);
+      if (F) {
```
vulder: same

include/vara/Feature/Relationship.h
```diff
@@ -0,0 +1,46 @@
...
+      : Kind(Kind), FeatureTreeNode(NodeKind::NK_RELATIONSHIP) {}
+
+  [[nodiscard]] RelationshipKind getKind() const { return Kind; }
+
+  static bool classof(const FeatureTreeNode *N) {
+    return N->getKind() == NodeKind::NK_RELATIONSHIP;
+  }
+
+private:
+  RelationshipKind Kind;
```
vulder:
```suggestion
  const RelationshipKind Kind;
```

include/vara/Feature/Relationship.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/SetVector.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/ADT/iterator_range.h"
+#include "llvm/IR/Function.h"
+#include "llvm/IR/Value.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <llvm/ADT/SetVector.h>
```
vulder:
```suggestion
#include "llvm/ADT/SetVector.h"

```
+ include sorting

include/vara/Feature/Relationship.h
```diff
@@ -0,0 +1,46 @@
...
+#include <set>
+#include <stack>
+#include <utility>
+#include <variant>
+
+using std::string;
+
```
vulder:
```suggestion
```

unittests/Feature/FeatureModelBuilder.cpp
```diff
@@ -34,81 +34,101 @@ TEST(FeatureModelBuilder, addOptionalFeature) {
   EXPECT_TRUE(FM->getFeature("a")->isOptional());
 }
 
-TEST(FeatureModelBuilder, addExclude) {
+// TODO(s9latimm): Replace string equals with subtree comparison
```
vulder: Should this be cleaned up?

unittests/Feature/FeatureModelBuilder.cpp
```diff
@@ -34,81 +34,101 @@ TEST(FeatureModelBuilder, addOptionalFeature) {
...
-
   B.makeFeature<BinaryFeature>("a");
-  B.addExclude("a", "b")->makeFeature<BinaryFeature>("b");
+  B.makeFeature<BinaryFeature>("b");
+  auto C = std::make_unique<ImpliesConstraint>(
+      std::make_unique<PrimaryFeatureConstraint>(
+          std::make_unique<Feature>("a")),
+      std::make_unique<NotConstraint>(
+          std::make_unique<PrimaryFeatureConstraint>(
+              std::make_unique<Feature>("b"))));
```
vulder, line 47: Seeing this multiple times, maybe it would make sense to create a ConstraintBuild but we can also do this later in a refactoring. If you want to do it later, could you please create an issue for it :smile:

unittests/Feature/FeatureModelWriter.cpp
```diff
@@ -37,7 +39,8 @@ TEST(XmlWriter, excludes) {
   EXPECT_FALSE(ActualOutput.empty());
 
   std::string ExpectedOutput = FS.get()->getBuffer();
-  EXPECT_EQ(ActualOutput, ExpectedOutput);
+  // TODO(s9latimm): Enable with refactored Constraint
+  //  EXPECT_EQ(ActualOutput, ExpectedOutput);
```
vulder: what is broken here?

unittests/Feature/FeatureModelWriter.cpp
```diff
@@ -54,7 +57,8 @@ TEST(XmlWriter, test) {
   FS = llvm::MemoryBuffer::getFileAsStream(getTestResource("test_out.xml"));
   EXPECT_TRUE(FS && "Comparisson file could not be read");
   std::string ExpectedOutput = FS.get()->getBuffer();
-  EXPECT_EQ(ActualOutput, ExpectedOutput);
+  // TODO(s9latimm): Enable with refactored Constraint
+  //  EXPECT_EQ(ActualOutput, ExpectedOutput);
```
vulder: same I guess?

vulder (changes requested):
So, I made a full pass over the code. This became quite a pull request :laughing: :+1:

unittests/Feature/FeatureModel.cpp
```diff
@@ -1,3 +1,4 @@
+#include "llvm/ADT/SetVector.h"
 #include <vara/Feature/FeatureModel.h>
```
vulder, line 2:
```suggestion
#include "vara/Feature/FeatureModel.h"

#include "llvm/ADT/SetVector.h"
```

include/vara/Feature/Feature.h
```diff
@@ -99,6 +105,57 @@ class Feature : public FeatureTreeNode {
...
+
+  llvm::iterator_range<excludes_iterator> excludes() {
+    std::vector<ExcludesConstraint *> Excludes;
+    for (auto *C : Constraints) {
+      auto *E = llvm::dyn_cast<ExcludesConstraint>(C);
+      if (E) {
+        Excludes.push_back(E);
+      }
+    }
+    return llvm::make_range(Excludes.begin(), Excludes.end());
```
vulder: You are returning a iterator to a local stack variable here, `Excludes` scope ends after this functions.
As far as I can tell, this function creates an arbitrary amount of nasal demons.
image

include/vara/Feature/Feature.h
```diff
@@ -99,6 +105,57 @@ class Feature : public FeatureTreeNode {
...
+  implications() const {
+    std::vector<ImpliesConstraint *> Implications;
+    for (auto *C : constraints()) {
+      auto *I = llvm::dyn_cast<ImpliesConstraint>(C);
+      if (I) {
+        Implications.push_back(I);
+      }
+    }
+    return llvm::make_range(Implications.begin(), Implications.end());
+  }
```
vulder: Same with these three cases.

include/vara/Feature/FeatureModel.h
```diff
@@ -154,9 +154,9 @@ class FeatureModelBuilder : private FeatureModel {
   }
 
   FeatureModelBuilder *
-  addRelationship(Relationship::RelationshipKind RK,
-                  const std::vector<std::string> &FeatureNames,
-                  const std::string &ParentName) {
+  emplaceRelationship(Relationship::RelationshipKind RK,
+                      const std::vector<std::string> &FeatureNames,
```
vulder, line 158:
```suggestion
                      std::vector<std::string> FeatureNames,
```

include/vara/Feature/FeatureModel.h
```diff
@@ -154,9 +154,9 @@ class FeatureModelBuilder : private FeatureModel {
...
+                      const std::string &ParentName) {
     RelationshipEdges[ParentName].emplace_back(RK, FeatureNames);
```
vulder, line 160:
```suggestion
    RelationshipEdges[ParentName].emplace_back(RK, std::move(FeatureNames));
```

vulder (changes requested):

include/vara/Feature/Constraint.h
```diff
@@ -19,7 +19,7 @@ namespace vara::feature {
 
 class ConstraintVisitor;
 
-/// \brief Tree like representation of constraints between features
+/// \brief Tree like representation of constraints between features3
```
```suggestion
/// \brief Tree like representation of constraints between features
```
I assume the `3` does not carry meaning :wink:

lib/Feature/FeatureModelWriter.cpp
```diff
@@ -184,12 +185,25 @@ int FeatureModelXmlWriter::writeBooleanConstraints( // NOLINT
...
+OrderedFeatureVector getChildrenFeatures(FeatureTreeNode *N) {
+  OrderedFeatureVector V;
+  for (FeatureTreeNode *C : N->children()) {
+    if (auto *F = llvm::dyn_cast<Feature>(C); F) {
+      V.insert(F);
+    } else {
+      V.insert(getChildrenFeatures(C));
+    }
+  }
+  return std::move(V);
```
vulder:
```suggestion
  return V;
```
If you return a local variable in all return paths C++17 guarantees return-value optimization, which can directly constructs the value in the caller stack frame. Using move here would prevent RVO.

lib/Feature/FeatureModelWriter.cpp
```diff
@@ -184,12 +185,25 @@ int FeatureModelXmlWriter::writeBooleanConstraints( // NOLINT
...
 
+OrderedFeatureVector getChildrenFeatures(FeatureTreeNode *N) {
+  OrderedFeatureVector V;
+  for (FeatureTreeNode *C : N->children()) {
+    if (auto *F = llvm::dyn_cast<Feature>(C); F) {
+      V.insert(F);
+    } else {
+      V.insert(getChildrenFeatures(C));
+    }
+  }
```
vulder: I would rewrite this a bit. Use a top level function with the same signature and a second private impl, which takes also a OrderedFeatureVector.
The top level function creates the OFV and passes it to the impl that fill the vector. With this impl you don't have to create a vector for every recursive step and delete it afterwards.
+ the interface for the user is the same :heavy_check_mark:

lib/Feature/FeatureModel.cpp
```diff
@@ -78,19 +78,44 @@ bool isSimpleMutex(const Feature *A, const Feature *B) {
       });
 }
 
+/// Collect mutual exclusive feature constraints for clean up.
+llvm::SmallSet<const Constraint *, 3>
+cleanUpMutex(const Feature *A, const llvm::SmallSet<Feature *, 3> &Xor) {
```
vulder: We could come up with a better name :laughing: My first thought was "why is he using a mutex here"
Maybe `cleanUpMutualExclusiveConstraints` ?

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -26,6 +26,12 @@ class OrderedFeatureVector {
   /// Insert feature while preserving ordering.
   void insert(Feature *F);
 
+  void insert(const OrderedFeatureVector &OFV) {
+    for (const auto &F : OFV) {
+      insert(F);
+    }
+  }
+
```
vulder: Together with the initializer list insert, we should create an insert that takes two iterators and just forward begin/end to it if we have a container insert like this or initializer list

vulder (changes requested):

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -13,40 +13,51 @@ namespace vara::feature {
...
-      typename std::vector<Feature *>::const_iterator;
+      typename llvm::SmallVector<Feature *, 3>::const_iterator;
 
   OrderedFeatureVector() = default;
+  OrderedFeatureVector(std::initializer_list<Feature *> Init) { insert(Init); }
   template <class FeatureIterTy>
-  OrderedFeatureVector(FeatureIterTy Start, FeatureIterTy End) {
-    insert(llvm::iterator_range(std::move(Start), std::move(End)));
+  OrderedFeatureVector(FeatureIterTy Begin, FeatureIterTy End) {
+    insert(std::move(Begin), std::move(End));
```
vulder:
```suggestion
    insert(std::forward<FeatureIterTy>(Begin), std::forward<FeatureIterTy>(End));
```
You need to use forward here because you don't not the type `FeatureIterTy`

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -13,40 +13,51 @@ namespace vara::feature {
...
       insert(F);
     }
   }
 
-  template <typename... Args> void insert(Feature *F, Args... FF) {
-    insert(F);
-    insert(FF...);
+  template <class FeatureIterTy>
+  void insert(FeatureIterTy Begin, FeatureIterTy End) {
+    insert(llvm::make_range(Begin, End));
```
vulder:
```suggestion
    insert(llvm::make_range(std::forward<FeatureIterTy>(Begin), std::forward<FeatureIterTy>(End)));
```

vulder (changes requested):

PR #25, opened 2020-11-10
Added validation and parsing of the xml part of sxfm

ChristianKaltenecker:


unittests/resources/test.sxfm
```diff
@@ -0,0 +1,53 @@
...
+		: compressionLevel_1(compressionLevel_1)
+	:g [1,1] processCount(processCount)
+		: processCount_64(processCount_64)
+		: processCount_32(processCount_32)
+		: processCount_16(processCount_16)
+		: processCount_8(processCount_8)
+	:g [1,1] mpm(mpm)
+		: worker(worker)
+		: event(event)
+	:o tls(tls)		:o aesTls(aesTls)			:g [0,1] Group_0(Group_0)
```
Kallistos: are multiple nodes in one line allowed?

unittests/resources/sxfm_example.sxfm
```diff
@@ -0,0 +1,51 @@
...
+  <data name="address">Model creator's address</data>
+  <data name="phone">Model creator's phone</data>
+  <data name="website">Model creator's website</data>
+  <data name="reference">Model's related publication</data>
+  </meta>
+  <feature_tree>
+    :r root (root_id)
+      :o opt1 (id_opt1)
+      :o opt2 (id_opt2)
+      :m man1
```
Kallistos, line 18: is it allowed that mandatory features have no (display-)name ?

unittests/resources/test.sxfm
ChristianKaltenecker: Nope. Seems to be a copy & paste error. This is fixed in the next revision. Thanks ;-)

unittests/resources/sxfm_example.sxfm
ChristianKaltenecker: If you are referring to the missing ID: yes, this is allowed. 
To be honest, I would simply enforce that the names have to be unique and give each of the features the id 'id_<name>'. This would simplify a lot...
@Kallistos What do you think?

Kallistos: That depends what the names of the features are.
If the feature name is the unique part right before the parenthesis, that is okay for me.
In SPLConqueror this is different right now. There the part in parenthesis is the name that is used as feature name.

ChristianKaltenecker: Correct, the feature name is the unique part right before the parenthesis. 
Yes, you are right regarding SPL Conqueror. However, I do not intend to implement is exactly as it is implemented in SPL Conqueror, as the implementation is partially error-prone.

ChristianKaltenecker: Btw. the mandatory feature above is from the original SXFM example of the SPLOT repository (http://www.splot-research.org/).

Kallistos: Okay!
Then the Id part in the parenthesis seems pretty useless anyways, right?

ChristianKaltenecker: Under the assumption that the feature names are already unique: yes, definitely!

include/vara/Feature/FeatureModelParser.h
```diff
@@ -90,9 +90,16 @@ class FeatureModelSxfmParser : public FeatureModelParser {
   std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> parseDoc();
   bool parseVm(xmlNode *Node);
 
-  static bool parseFeatureTree(xmlChar *FeatureTree);
+  bool parseFeatureTree(xmlChar *FeatureTree);
   static bool parseConstraints(xmlChar *Constraints);
 
+  static int countOccurrences(const string& Target, const string& StringToSearch);
+  static std::optional<std::tuple<string, string>> extractCardinality(const string& StringToExtractFrom);
```
Kallistos: why are the cardinalities represented as strings instead of ints, or Integer?

include/vara/Feature/FeatureModelParser.h
```diff
@@ -90,9 +90,16 @@ class FeatureModelSxfmParser : public FeatureModelParser {
...
+  static string readUntil(const string& StringToReadFrom, const char& CharToSearch, std::string::size_type Start);
+  bool
+  processFeatureTreeLine(std::stringstream& Ss, const char *ParentName,
+                         std::optional<std::tuple<string, string>> ParentCardinality,
```
Kallistos: same as above

lib/Feature/FeatureModelParser.cpp
```diff
@@ -249,29 +250,242 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlChar *FeatureTree) {
   {
     std::stringstream Ss(reinterpret_cast<const char *>(FeatureTree));
     std::string To;
+    string Name;
+    bool Opt;
+    int LastIndentation = -1;
+    int OrGroupCounter = 0;
+    std::map<int, string> IndentationToParentMapping;
```
Kallistos, line 363: Why do you remember different indentations ? and what happens if you have two indentations that are on the same level but have different parents?

Isn't the current indentation level sufficient?

lib/Feature/FeatureModelParser.cpp
```diff
@@ -249,29 +250,242 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlChar *FeatureTree) {
...
+        continue;
+      }
 
+      // For every line, count the indentation
+      // -1,0 or +1 additional indentations are allowed to the original one
+      // Otherwise, the format is violated
+      int CurrentIndentationLevel = countOccurrences(To, Indentation);
+      int Diff = LastIndentation - CurrentIndentationLevel;
+      if ((LastIndentation != -1) && (Diff < -1 | Diff > 1)) {
+        std::cerr << "Indentation error in feature tree." << std::endl;
```
Kallistos: You could add more specific information about the indentation: what you expected and what you got.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -249,29 +250,242 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlChar *FeatureTree) {
...
+        return false;
       }
+
+      // Move pointer to first character after indentation
+      // The first character has to be a colon followed by the type of
+      // the feature (m for mandatory, o for optional, a for alternative)
+      std::string::size_type Pos = CurrentIndentationLevel * Indentation.length() + 2;
+      std::optional<std::tuple<string, string>> Cardinalities;
+      switch (To.at(Pos - 1)) {
```
Kallistos, line 412: There is a check for the colon missing.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -249,29 +250,242 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlChar *FeatureTree) {
...
+      // Remove the cardinality
+      if (Name.find_first_of('[') != std::string::npos) {
+        Name = readUntil(Name, '[', 0);
+      }
+
+      // If there is no name, provide an artificial one
+      if (Name.empty()) {
+        OrGroupCounter++;
+        Name = "group_" + std::to_string(OrGroupCounter);
+      }
```
Kallistos: after you checked the indentation you could just split the input line at whitespaces 

`std::vector<std::string> tokens;`
`split(To, " ", std::back_inserter(tokens));`

lib/Feature/FeatureModelParser.cpp
```diff
@@ -249,29 +250,242 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlChar *FeatureTree) {
...
+  // Note that we ignore the ID and use the name of the feature
+  // as unique identifier.
+  if (Name.find_first_of('(') != std::string::npos) {
+    Name = readUntil(Name, '(', 0);
+  }
+
+  // Remove the cardinality
+  if (Name.find_first_of('[') != std::string::npos) {
+    Name = readUntil(Name, '[', 0);
+  }
```
Kallistos: That is the same code as at line 277, why did you wrote that twice?

lib/Feature/FeatureModelParser.cpp
```diff
@@ -249,29 +250,242 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlChar *FeatureTree) {
...
+  if (Pos > StringToReadFrom.length()) {
+    std::cerr << "Position out of bounds!" << std::endl;
+    return "";
+  }
+
+  while ((Pos < StringToReadFrom.length()) && (StringToReadFrom.at(Pos) != CharToSearch)) {
+    Result += StringToReadFrom.at(Pos);
+    Pos++;
+  }
+  return Result;
```
Kallistos: you might not need that method if you use the split method on strings

include/vara/Feature/FeatureModelParser.h
ChristianKaltenecker: Because there is also the '*' sign for the wildcard. Other than that, an integer would suffice.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: Ah, I forgot to remove this could after I changed my approach from a recursive way to a loop.
Will be removed with the next commit.

include/vara/Feature/FeatureModelParser.h
Kallistos: You could use INT_MAX for '*'

ChristianKaltenecker: I would rather prefer '-1' because '-1' as indentation is not a possible value ;-)
Or, because I have no other negative values, I could also use an unsigned int and use INT_MAX for '*'.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: We do have to remember the different indentations for cases such as:
```
:r root (root_id)
  :o opt1 (id_opt1)
    :o opt2 (id_opt2)
      :o opt3 (id_opt3)
  :o opt4 (id_opt4)
```
The indentation between op3 and opt4 is -2 and the parent of opt4 is root and not opt2.
Thus, it does not suffice to only remember the last parent. 

> and what happens if you have two indentations that are on the same level but have different parents?

Should be no problem since a new parent of a certain indentation level overwrites the last one. There is no case where we would still need the last parent.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: Sure. I added more comments and also found a small bug in line 279. 
The bug is now also covered in a test.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: Correct. Will be included in the next commit.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+      std::cerr << "Failed to validate DTD." << std::endl;
+    }
+  } else {
+    std::cerr << "Failed to parse / validate XML." << std::endl;
+  }
+  return std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)>(nullptr, nullptr);
+}
+
+/// This method checks if the given feature model is valid
+/// \return true iff the feature model is valid
```
```suggestion
/// This method checks if the given feature model is valid
///
/// \returns true iff the feature model is valid
```

And please move the docs to the header file.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+bool FeatureModelSxfmParser::verifyFeatureModel() { return parseDoc().get(); }
```
vulder: Could be an inline function in the header.

include/vara/Feature/FeatureModelParser.h
```diff
@@ -64,6 +64,41 @@ class FeatureModelXmlParser : public FeatureModelParser {
...
 };
 
+//===----------------------------------------------------------------------===//
+//                         FeatureModelSxfmParser Class
+//===----------------------------------------------------------------------===//
+
+/// \brief Parsers for feature models in SXFM.
+/// The SXFM (simple XML feature model) format is a feature model format embeded
+/// in an XML structure.
+class FeatureModelSxfmParser : public FeatureModelParser {
```
vulder, line 74: Could you in general, please try to order the functions in the cpp file as they are in the header file :smile:. Makes it easier to find.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+
+  return Result;
+}
+
+/// This method reads beginning from a certain starting position and reads in
+/// every character until CharToSearch is found.
+/// \param StringToReadFrom string to read from
+/// \param CharToSearch character to search for
+/// \param Start the starting position to start reading
+/// \return the string read until the given character is found
```
```suggestion
/// This method reads beginning from a certain starting position and reads in
/// every character until CharToSearch is found.
///
/// \param StringToReadFrom string to read from
/// \param CharToSearch character to search for
/// \param Start the starting position to start reading
///
/// \returns the string read until the given character is found
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+string FeatureModelSxfmParser::readUntil(const string& StringToReadFrom, const char& CharToSearch, std::string::size_type Start) {
```
vulder: You can just use `llvm::StringRef::find`

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  return std::optional<std::tuple<int, int>>(std::tuple<int, int>{MinCardinality.value(), MaxCardinality.value()});
+}
+
+
+/// This method parses the given cardinality and returns an optional.
+/// If the optional is empty, the process failed; otherwise the result contains
+/// either UINT_MAX for the wildcard or the cardinality number as integer.
+/// \param CardinalityString the cardinality to parse
+/// \return an optional that contains no integer in case of failure or UINT_MAX
+/// for wildcard, or the number itself.
```
```suggestion
/// This method parses the given cardinality and returns an optional.
/// If the optional is empty, the process failed; otherwise the result contains
/// either UINT_MAX for the wildcard or the cardinality number as integer.
///
/// \param CardinalityString the cardinality to parse
///
/// \returns the parsed number, UINT_MAX for wildcards, and nullopt in case of a failure
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+std::optional<int> FeatureModelSxfmParser::parseCardinality(const string& CardinalityString) {
+  std::optional<int> Result = std::optional<int>();
```
vulder, line 538:
```suggestion
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  if (CardinalityString == "*") {
+    // We use UINT_MAX as our magic integer (which is -1 as int) to indicate
+    // that the cardinality is a wildcard.
+    Result = UINT_MAX;
```
vulder, line 542:
```suggestion
    return UINT_MAX;
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+    LongNumber = strtol(CardinalityString.c_str(), &End, 0);
+    if (errno == ERANGE || LongNumber < INT_MIN || LongNumber > INT_MAX
+        || *CardinalityString.c_str() == '\0' || *End != '\0') {
+      std::cerr << "The following cardinality is not integer: " << CardinalityString << std::endl;
+    } else {
+      Result = LongNumber;
+    }
+  }
+
+  return Result;
```
vulder, line 557:
```suggestion
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+    Result = UINT_MAX;
+  } else {
+    // Convert the string into an integer in a safe way
+    char *End;
+    long LongNumber;
+    errno = 0;
+    LongNumber = strtol(CardinalityString.c_str(), &End, 0);
+    if (errno == ERANGE || LongNumber < INT_MIN || LongNumber > INT_MAX
+        || *CardinalityString.c_str() == '\0' || *End != '\0') {
+      std::cerr << "The following cardinality is not integer: " << CardinalityString << std::endl;
```
vulder, line 551:
```suggestion
      llvm::errs() << "The following cardinality is not integer: " << CardinalityString << '\n';
```
We try to use the llvm out and error stream because they are a bit easier to use and have some interesting features. So, to be consistent, it would be good to use them everywhere.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+    } else {
```
vulder, line 552:
```suggestion
      return std::nullopt;
    } else {
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+}
+
+
+/// This method parses the given cardinality and returns an optional.
+/// If the optional is empty, the process failed; otherwise the result contains
+/// either UINT_MAX for the wildcard or the cardinality number as integer.
+/// \param CardinalityString the cardinality to parse
+/// \return an optional that contains no integer in case of failure or UINT_MAX
+/// for wildcard, or the number itself.
+std::optional<int> FeatureModelSxfmParser::parseCardinality(const string& CardinalityString) {
```
vulder, line 537: If you only read a string it is good to only use a `std::string_view`, for llvm there is a similar type called `llvm::StringRef`. Try to use a `StringRef` here, makes it easier because there is already a function `getAsInteger` :smile:

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+    // Convert the string into an integer in a safe way
+    char *End;
+    long LongNumber;
+    errno = 0;
+    LongNumber = strtol(CardinalityString.c_str(), &End, 0);
+    if (errno == ERANGE || LongNumber < INT_MIN || LongNumber > INT_MAX
+        || *CardinalityString.c_str() == '\0' || *End != '\0') {
+      std::cerr << "The following cardinality is not integer: " << CardinalityString << std::endl;
+    } else {
+      Result = LongNumber;
```
vulder, line 553:
```suggestion
      return LongNumber;
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  }
+
+  return true;
+}
+
+/// This method extracts the cardinality from the given line.
+/// The cardinality is wrapped in square brackets (e.g., [1,1])
+/// \param StringToExtractFrom the string to extract the cardinality from
+/// \return
+std::optional<std::tuple<int, int>> FeatureModelSxfmParser::extractCardinality(const string& StringToExtractFrom) {
```
vulder, line 507:
```suggestion
std::optional<std::tuple<int, int>> FeatureModelSxfmParser::extractCardinality(const string& StringToExtractFrom) {
```
Can take a StringRef, see description bellow.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  std::optional<int> MinCardinality;
+  std::optional<int> MaxCardinality;
+
```
vulder, line 510: With StringRefs all the positon/string handling becomes much easier.

See bellow

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+                                                  SxfmConstants::DtdRaw.length(),
+                                                  XML_CHAR_ENCODING_UTF8),
+                    XML_CHAR_ENCODING_UTF8),
+      xmlFreeDtd);
+  xmlCleanupParser();
+  assert(Dtd && "Failed to parse DTD.");
+  return Dtd;
+}
+
+bool FeatureModelSxfmParser::parseFeatureTree(xmlChar *FeatureTree) {
```
vulder, line 353: This function is quite huge and should be split into different helper functions.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+    int OrGroupCounter = 0;
+    std::map<int, string> IndentationToParentMapping;
+
+    // This map is used for the or group mapping
+    // Each entry represents an or group as a tuple where the first value is
+    // the name of the parent, the second is the relationship kind, and the
+    // third a vector consisting of the name of the children
+    std::map<int, std::tuple<string, Relationship::RelationshipKind, std::vector<string>>> OrGroupMapping;
+
+    if (FeatureTree == nullptr) {
```
vulder, line 371:
```suggestion
    if (!FeatureTree) {
```
Could also be

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+
+      // Move pointer to first character after indentation
+      // The first character has to be a colon followed by the type of
+      // the feature (m for mandatory, o for optional, a for alternative)
+      std::string::size_type Pos = CurrentIndentationLevel * Indentation.length() + 2;
+      std::optional<std::tuple<int, int>> Cardinalities;
+
+      switch (To.at(Pos - 1)) {
+      case 'o':
+        // Code for optional
```
vulder, line 414: If these code have special meaning, would it make sense to create a small enum that make the mapping clear? So we could write:
```
case SXFM_CHAR_CODE::OPTIONAL:
    // ... do stuff
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+      if ((LastIndentationLevel != -1) && Diff > 1) {
+        std::cerr << "Indentation error in feature tree in line " << To << std::endl;
+        return false;
+      }
+
+      // Move pointer to first character after indentation
+      // The first character has to be a colon followed by the type of
+      // the feature (m for mandatory, o for optional, a for alternative)
+      std::string::size_type Pos = CurrentIndentationLevel * Indentation.length() + 2;
+      std::optional<std::tuple<int, int>> Cardinalities;
```
vulder, line 410: In `cpp` files it can be helpful to add using decl for specific names that are used often and don't collide.
E.g.:
```
using std::optional;

// ... code
```
```suggestion
      optional<std::tuple<int, int>> Cardinalities;
```

PS: there is also using namespace, however, I would recommend to use single using decls because it makes it clearer at the top of the file what names are importet. Furthermore, it reduces the chance of unwanted name clashes, especially if frameworks/stdlib or otherthings are updated.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+
+      // Add parent from the upper indentation level if there is one
+      if (LastIndentationLevel != -1 && CurrentIndentationLevel == RootIndentation) {
+        std::cerr << "Only one feature can be root and have the same indentation as root." << std::endl;
+        return false;
+      }
+
+      if (LastIndentationLevel != -1) {
+        auto Parent = IndentationToParentMapping.find(CurrentIndentationLevel - 1);
+        assert(Parent != IndentationToParentMapping.end());
```
vulder, line 463: Hmm, is this more the case that "this should normally never happen" or "a wrong formatted file causes this case". In case of the second one, I would suggest different error handling because if this case happens in an optimized build -- no asserts -- the dereference bellow is UB and could/will crashe hard.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+      if (OrGroup != OrGroupMapping.end()) {
+        std::get<2>(OrGroup->second).push_back(Name);
+      }
+
+
+      LastIndentationLevel = CurrentIndentationLevel;
+    }
+
+    // Add the remaining or groups
+    for (auto & OrGroup : OrGroupMapping) {
```
vulder, line 494:
```suggestion
    for (const auto &OrGroup : OrGroupMapping) {
```
Formatting + const.
`OrGroup` is never modified in the loop and can, therefore, be const.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  return Result;
+}
+
+/// This method counts the occurrences of the second argument in the first argument.
+/// The intended use of this method is to count the indentation level of the feature
+/// tree in the sxfm format.
+/// \param StringToSearch the string to search in
+/// \param StringToFind the string to count the occurrences of
+/// \return the number of occurrences
+int FeatureModelSxfmParser::countOccurrences(const string& StringToSearch, const string& StringToFind) {
```
```suggestion
int FeatureModelSxfmParser::countOccurrences(const string& StringToSearch, const string& StringToFind) {
```
Has different parameter names in the header file.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  }
+  return Result;
+}
+
+/// This method counts the occurrences of the second argument in the first argument.
+/// The intended use of this method is to count the indentation level of the feature
+/// tree in the sxfm format.
+/// \param StringToSearch the string to search in
+/// \param StringToFind the string to count the occurrences of
+/// \return the number of occurrences
```
vulder: Move to header.

```suggestion
/// This method counts the occurrences of the second argument in the first argument.
/// The intended use of this method is to count the indentation level of the feature
/// tree in the sxfm format.
///
/// \param StringToSearch the string to search in
/// \param StringToFind the string to count the occurrences of
///
/// \returns number of occurrences
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  while ((Pos = StringToSearch.find(StringToFind, Pos )) != std::string::npos) {
+    ++Occurrences;
+    Pos += StringToFind.length();
+  }
+  return Occurrences;
+}
+
+bool FeatureModelSxfmParser::parseConstraints(xmlChar *Constraints) {
+  // TODO (CK): This has to wait until the constraint part is implemented in
+  // the FeatureModelBuilder (issue #24)
```
```suggestion
  // TODO (se-passau/VaRA#$NUMBER): This has to wait until the constraint part is implemented in
  // the FeatureModelBuilder
```
The issue number `issue #24` is not in the issue tracker, did you miss typed this?

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  return true;
+}
+
+bool FeatureModelSxfmParser::parseVm(xmlNode *Node) {
+  // Parse the name first. This procedure is the same as for parsing the
+  // XML file.
```
vulder: a) what do you mean by "the same as for parsing the xml file" Is sxfm not also an Xml file?
b) If it is the same we could factor our a method that does the work for both

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  {
+    std::unique_ptr<xmlChar, void (*)(void *)> Cnt(xmlGetProp(Node, XmlConstants::NAME),
+                                                   xmlFree);
```
vulder: Can this fail and initialize `Cnt` with a nullptr?

lib/Feature/FeatureModelParser.cpp
```diff
@@ -225,4 +252,361 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+      if (!xmlStrcmp(H->name, SxfmConstants::FEATURE_TREE)) {
+        // Check whether a custom indentation is defined in the feature tree and
+        // set it accordingly
+        {
+          std::unique_ptr<xmlChar, void (*)(void *)> Cnt(xmlGetProp(H, SxfmConstants::INDENTATION),
+                                                         xmlFree);
+          if (Cnt) {
+            Indentation = reinterpret_cast<char*>(Cnt.get());
+          }
+
```
```suggestion
```

vulder (changes requested):
I did a first pass over the changes, if you have questions just ping me :smiley:

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: I have adjusted this comment, since there are some differences to the way how xml files are parsed.
I see little reason to refactor here.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: No, this shouldn't happen, since I demand the name attribute to be included in the sxfm model. If it is not available, the xml verification would actually fail **before** it is even parsed and processed.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: Yes, I misstyped this.I have changed the reference to issue `#702`, which is the right one.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: I have done that now. I could also remove another method, which counted the occurrences of a certain string in another one.

lib/Feature/FeatureModelParser.cpp
vulder: k

lib/Feature/FeatureModelParser.cpp
vulder: k

lib/Feature/FeatureModelParser.cpp
vulder: thx

ChristianKaltenecker:
This will be addressed in another PR.

PR #29, opened 2020-12-11
SXFM parser

ChristianKaltenecker:


lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  // In the following, the document is validated.
+  // Therefore, (1) check whether it could be parsed
+  if (Doc && Ctxt->valid) {
+    // (2) validate the sxfm format by using the dtd (document type definition) file
+    if (xmlValidateDtd(&Ctxt->vctxt, Doc.get(), createDtd().get())) {
+      // and (3) check the tree-like structure of the embedded feature model
+      // as well as constraints
+      return Doc;
+    } else {
+      std::cerr << "Failed to validate DTD." << std::endl;
```
```suggestion
      llvm::errs() << "Failed to validate DTD.\n";
```
Please try to use llvm streams because they work better together with other llvm string formating etc.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -411,15 +411,26 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlChar *FeatureTree) {
...
+          if (!Tokens.second.empty()) {
+            Name = Tokens.second.str();
+          }
+        }
+      }
+
+      // Note that we ignore the ID and use the name of the feature
+      // as unique identifier.
+      if (Name.find_first_of('(') != std::string::npos) {
+        Name = Name.substr(0, Name.find('(', 0));
```
Kallistos, line 500: shouldn't ')' mark the end of the name?

unittests/Feature/FeatureModelSxfm.cpp
```diff
@@ -12,17 +12,33 @@ namespace vara::feature {
 TEST(SxfmParser, validation) {
```
Kallistos, line 17: Maybe add a short description to each test what the test wants to test

include/vara/Feature/FeatureModelParser.h
```diff
@@ -64,6 +64,89 @@ class FeatureModelXmlParser : public FeatureModelParser {
...
 
+//===----------------------------------------------------------------------===//
+//                         FeatureModelSxfmParser Class
+//===----------------------------------------------------------------------===//
+
+/// \brief Parsers for feature models in SXFM.
+/// The SXFM (simple XML feature model) format is a feature model format embeded
+/// in an XML structure.
+class FeatureModelSxfmParser : public FeatureModelParser {
+public:
```
vulder, line 78:
```suggestion
public:
  using UniqueXmlDoc = std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)>;
  using UniqueXmlDtd = std::unique_ptr<xmlDtd, void (*)(xmlDtdPtr)>;

```
Hmm, I think we could add some short cuts for the xml unique_ptrs :)

include/vara/Feature/FeatureModelParser.h
```diff
@@ -64,6 +64,89 @@ class FeatureModelXmlParser : public FeatureModelParser {
...
+
+  std::string Sxfm;
+  FeatureModelBuilder FMB;
+  std::string Indentation = "\t";
+
+  /// Returns a pointer to the dtd representation of the xml file, which
+  /// is needed to verify the structure of the xml file.
+  ///
+  /// \returns a pointer to the dtd representation
+  static std::unique_ptr<xmlDtd, void (*)(xmlDtdPtr)> createDtd();
```
vulder:
```suggestion
  static UniqueXmlDtd createDtd();
```

include/vara/Feature/FeatureModelParser.h
```diff
@@ -64,6 +64,89 @@ class FeatureModelXmlParser : public FeatureModelParser {
...
+
+  /// Parses the given xml file by using libxml2 and returns a pointer to
+  /// the xml document.
+  ///
+  /// \returns a pointer to the xml document
+  std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> parseDoc();
```
vulder:
```suggestion
  UniqueXmlDoc parseDoc();
```

include/vara/Feature/FeatureModelParser.h
```diff
@@ -64,6 +64,89 @@ class FeatureModelXmlParser : public FeatureModelParser {
...
+
+  /// This method parses the given cardinality and returns an optional.
+  /// If the optional is empty, the process failed; otherwise the result contains
+  /// either UINT_MAX for the wildcard or the cardinality number as integer.
+  ///
+  /// \param CardinalityString the cardinality to parse
+  ///
+  /// \returns an optional that contains no integer in case of failure or UINT_MAX
+  /// for wildcard, or the number itself.
+  static std::optional<int> parseCardinality(const string& CardinalityString);
```
vulder:
```suggestion
  static std::optional<int> parseCardinality(const string& CardinalityString);

  std::string Sxfm;
  FeatureModelBuilder FMB;
  std::string Indentation = "\t";
```

Currently, we try to put members last, to follow "some" style guide (mostly google).

include/vara/Feature/FeatureModelParser.h
```diff
@@ -64,6 +64,89 @@ class FeatureModelXmlParser : public FeatureModelParser {
...
+  ///
+  /// \returns the feature model that was read in
+  std::unique_ptr<FeatureModel> buildFeatureModel() override;
+
+private:
+  using constSxfmCharPtr = const xmlChar *;
+
+  std::string Sxfm;
+  FeatureModelBuilder FMB;
+  std::string Indentation = "\t";
```
vulder:
```suggestion
```

lib/Feature/FeatureModel.cpp
```diff
@@ -109,7 +109,7 @@ void FeatureModelBuilder::detectXMLAlternatives() {
     std::vector<std::string> Frontier(Children[FeatureName].begin(),
                                       Children[FeatureName].end());
     while (!Frontier.empty()) {
-      const auto &FName = Frontier.back();
+      const auto &FName = std::string(Frontier.back());
```
vulder:
```suggestion
      const auto FName = std::string(Frontier.back());
```
We should use a value here, with a ref you get a reference to a temporary, which can be dangerous, for example, when someone removes the const because only const references extend the lifetime of a temporary.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
 // TODO(s9latimm): replace with builder err
 bool FeatureModelXmlParser::verifyFeatureModel() { return parseDoc().get(); }
 
+//===----------------------------------------------------------------------===//
+//                        FeatureModelSxfmParser Class
+//===----------------------------------------------------------------------===//
+
+std::unique_ptr<FeatureModel> FeatureModelSxfmParser::buildFeatureModel() {
+  std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> Doc = parseDoc();
```
```suggestion
  UniqueXmlDoc Doc = parseDoc();
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  if (!Doc) {
+    return nullptr;
+  }
+
+  FMB.init();
+  return parseVm(xmlDocGetRootElement(Doc.get())) ? FMB.buildFeatureModel()
+                                                  : nullptr;
+}
+
+std::unique_ptr<xmlDtd, void (*)(xmlDtdPtr)>
```
```suggestion
UniqueXmlDtd
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+FeatureModelSxfmParser::createDtd() {
+  std::unique_ptr<xmlDtd, void (*)(xmlDtdPtr)> Dtd(
```
```suggestion
  UniqueXmlDtd Dtd(
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+
+  // In the following, the document is validated.
+  // Therefore, (1) check whether it could be parsed
+  if (Doc && Ctxt->valid) {
+    // (2) validate the sxfm format by using the dtd (document type definition) file
+    if (xmlValidateDtd(&Ctxt->vctxt, Doc.get(), createDtd().get())) {
+      // and (3) check the tree-like structure of the embedded feature model
+      // as well as constraints
+      return Doc;
+    } else {
```
```suggestion
    }
```
You don't need an `else` when you have a return in the then case.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+      std::cerr << "Failed to validate DTD." << std::endl;
+    }
```
vulder, line 331:
```suggestion
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  } else {
+    std::cerr << "Failed to parse / validate XML." << std::endl;
```
```suggestion
    llvm::errs() << "Failed to parse / validate XML.\n";
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  }
+  return std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)>(nullptr, nullptr);
```
```suggestion
  return UniqueXmlDoc(nullptr, nullptr);
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+}
+
+bool FeatureModelSxfmParser::parseVm(xmlNode *Node) {
+  // Parse the name first.
+  {
+    std::unique_ptr<xmlChar, void (*)(void *)> Cnt(xmlGetProp(Node, XmlConstants::NAME),
```
vulder: Maybe we should also create a `UniqueXmlChar` using decl.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+
+      if (To.empty() || std::all_of(To.begin(),To.end(),isspace)) {
+        continue;
+      }
+
+      // For every line, count the indentation
+      // not more than 1 additional indentations are allowed to the original one
+      // However, we may have arbitrarily less indentations
+      if ((To.find(':', 0 )) == std::string::npos) {
+        std::cerr << "Colon is missing in line" << To << std::endl;
```
vulder: llvm streams

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+    std::map<int, string> IndentationToParentMapping;
+
+    // This map is used for the or group mapping
+    // Each entry represents an or group as a tuple where the first value is
+    // the name of the parent, the second is the relationship kind, and the
+    // third a vector consisting of the name of the children
+    std::map<int, std::tuple<string, Relationship::RelationshipKind, std::vector<string>>> OrGroupMapping;
+
+    if (FeatureTree == nullptr) {
+      std::cerr << "Failed to read in feature tree. Is it empty?" << std::endl;
```
vulder: llvm streams

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+
+      // For every line, count the indentation
+      // not more than 1 additional indentations are allowed to the original one
+      // However, we may have arbitrarily less indentations
+      if ((To.find(':', 0 )) == std::string::npos) {
+        std::cerr << "Colon is missing in line" << To << std::endl;
+        return false;
+      }
+
+      llvm::StringRef ToStringRef = llvm::StringRef(To);
```
```suggestion
      llvm::StringRef ToStringRef(To);
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+      int CurrentIndentationLevel = IndentationString.count(Indentation);
+      int Diff = CurrentIndentationLevel - LastIndentationLevel;
+
+      // Remember the root indentation for later checks
+      if (LastIndentationLevel == -1) {
+        RootIndentation = CurrentIndentationLevel;
+      }
+
+      if ((LastIndentationLevel != -1) && Diff > 1) {
+        std::cerr << "Indentation error in feature tree in line " << To << std::endl;
```
vulder: llvm streams

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+        Cardinalities = extractCardinality(To);
+        if (!Cardinalities.has_value()) {
+          return false;
+        }
+        break;
+      case ' ':
+        // Code for alternative child
+        Pos--;
+        break;
+      }
```
vulder, line 472: What happens here if we get a different letter? Does the Dtd prevent this, so can we be sure that we won't see anything else except for these three, or should we add some `default` label with error handling here?

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+        Name = "group_" + std::to_string(OrGroupCounter);
+      }
+
+      // Create the feature
+      FMB.makeFeature<BinaryFeature>(Name, Opt);
+      IndentationToParentMapping[CurrentIndentationLevel] = Name;
+
+      // Add parent from the upper indentation level if there is one
+      if (LastIndentationLevel != -1 && CurrentIndentationLevel == RootIndentation) {
+        std::cerr << "Only one feature can be root and have the same indentation as root." << std::endl;
```
vulder: llvm streams

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+      if (LastIndentationLevel != -1) {
+        auto Parent = IndentationToParentMapping.find(CurrentIndentationLevel - 1);
+        assert(Parent != IndentationToParentMapping.end());
+        FMB.addParent(Name, Parent->second);
+      }
+
+      // Add the or group to the feature model if it is completely parsed
+      auto OrGroup = OrGroupMapping.find(CurrentIndentationLevel);
+      if (OrGroup != OrGroupMapping.end()) {
+        FMB.emplaceRelationship(std::get<1>(OrGroup->second), std::get<2>(OrGroup->second), std::get<0>(OrGroup->second));
```
vulder: formatting, this line is obviously to to long. If you need help setting up clang-format just ping me

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+
+std::optional<std::tuple<int, int>> FeatureModelSxfmParser::extractCardinality(const string& StringToExtractFrom) {
+  std::optional<int> MinCardinality;
+  std::optional<int> MaxCardinality;
+
+  // Search for the first occurrence of '['; then read in the min cardinality
+  // until the comma. Afterwards, read in the max cardinality until ']'
+  std::string::size_type Pos = StringToExtractFrom.find_first_of('[');
+  if (Pos == std::string::npos) {
+    std::cerr << "No cardinality given in or group!" << std::endl;
```
vulder: llvm streams

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+    return std::optional<std::tuple<int, int>>();
+  }
+  auto CardinalityString = llvm::StringRef(StringToExtractFrom);
```
```suggestion
  llvm::StringRef CardinalityString(StringToExtractFrom);
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  size_t CommaPos = CardinalityString.find(',', Pos + 1);
+  MinCardinality = parseCardinality(CardinalityString.substr(Pos + 1, CommaPos - Pos - 1).str());
+  Pos = CommaPos;
+  MaxCardinality = parseCardinality(CardinalityString.substr(Pos + 1, CardinalityString.find(']', Pos + 1) - Pos - 1).str());
+
+  if (!MinCardinality.has_value() || !MaxCardinality.has_value()) {
+    std::cerr << "No parsable cardinality!" << std::endl;
```
vulder: llvm streams

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+    return std::optional<std::tuple<int, int>>();
+  }
+
+  if (MinCardinality.value() != 1 || (MaxCardinality.value() != 1 && MaxCardinality.value() != UINT_MAX)) {
+    std::cerr << "Cardinality unsupported. We support cardinalities [1,1] (alternative) or [1, *] (or group)." << std::endl;
```
vulder: llvm streams

include/vara/Feature/FeatureModelParser.h
vulder:
```suggestion
  static std::optional<int> parseCardinality(llvm::StringRef CardinalityString);
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+    return std::optional<std::tuple<int, int>>();
+  }
+
+  return std::optional<std::tuple<int, int>>(std::tuple<int, int>{MinCardinality.value(), MaxCardinality.value()});
+}
+
+std::optional<int> FeatureModelSxfmParser::parseCardinality(const string& CardinalityString) {
```
```suggestion
std::optional<int> FeatureModelSxfmParser::parseCardinality(llvm::StringRef CardinalityString) {
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  std::optional<int> Result = std::optional<int>();
+  if (CardinalityString == "*") {
+    // We use UINT_MAX as our magic integer (which is -1 as int) to indicate
+    // that the cardinality is a wildcard.
+    Result = UINT_MAX;
+  } else {
+    // Convert the string into an integer in a safe way
+    char *End;
```
```suggestion
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -233,4 +234,318 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+    long LongNumber;
+    errno = 0;
+    LongNumber = strtol(CardinalityString.c_str(), &End, 0);
+    if (errno == ERANGE || LongNumber < INT_MIN || LongNumber > INT_MAX
+        || *CardinalityString.c_str() == '\0' || *End != '\0') {
+      std::cerr << "The following cardinality is not integer: " << CardinalityString << std::endl;
```
```suggestion
    if (!llvm::to_integer(CardinalityString, LongNumber, 10)) {
      llvm::errs() << llvm::formatv("The cardinality: '{0}' was not an integer.\n", CardinalityString);
```
llvm has us already covered to make converting strings to numbers easier :smiley:

lib/Feature/FeatureModelParser.cpp
```diff
@@ -1,6 +1,7 @@
 #include "vara/Feature/FeatureModelParser.h"
 
 #include "XmlConstants.h"
+#include "SxfmConstants.h"
```
```suggestion

#include "llvm/ADT/StringExtras.h"
#include "llvm/Support/FormatVariadic.h"

#include "SxfmConstants.h"
#include "XmlConstants.h"
```
Fixes include ordering and adds includes needed for other suggestions further down.

vulder (dismissed):
Looks good, with this we should finally be able to sparse sxfm files :+1: 
Found just a few things to clean up, otherwise, this looks fine to me.

PS: after applying the suggestions you should run a formatter, gh can not do this automatically :laughing:
or in general run a formatter^^

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: As far as I know, this marks the end of the ID. And the ID is not always included. That is also the reason why I am currently ignoring it.

unittests/Feature/FeatureModelSxfm.cpp
ChristianKaltenecker: This is done in the next commit. Thanks!

include/vara/Feature/FeatureModelParser.h
ChristianKaltenecker: This should be fixed now. Could you check it, please?

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: Please resolve this conversation if this is okay for you ;-)

include/vara/Feature/FeatureModelParser.h
ChristianKaltenecker: Maybe it is also a good idea to shift this into the superclass FeatureModelParser itself since the XML implementation could use these shortcuts too.

vulder: If it's in general the same that sounds like a good idea.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: You are right. I definitely missed that. Thanks!

lib/Feature/FeatureModel.cpp
```diff
@@ -173,7 +173,7 @@ void FeatureModelBuilder::detectXMLAlternatives() {
     std::vector<std::string> Frontier(Children[FeatureName].begin(),
                                       Children[FeatureName].end());
     while (!Frontier.empty()) {
-      const auto &FName = Frontier.back();
+      const auto FName = std::string(Frontier.back());
```
vulder:
```suggestion
      const std::string FName{Frontier.back()};
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -21,10 +25,8 @@ bool FeatureModelXmlParser::parseConfigurationOption(xmlNode *Node,
   std::vector<FeatureSourceRange> SourceRanges;
   for (xmlNode *Head = Node->children; Head; Head = Head->next) {
     if (Head->type == XML_ELEMENT_NODE) {
-      string Cnt = std::string(
-          reinterpret_cast<char *>(std::unique_ptr<xmlChar, void (*)(void *)>(
-                                       xmlNodeGetContent(Head), xmlFree)
-                                       .get()));
+      string Cnt = std::string(reinterpret_cast<char *>(
+          UniqueXmlChar(xmlNodeGetContent(Head), xmlFree).get()));
```
```suggestion
      std::string Cnt{reinterpret_cast<char *>(
          UniqueXmlChar(xmlNodeGetContent(Head), xmlFree).get())};
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -151,10 +151,8 @@ bool FeatureModelXmlParser::parseConstraints(xmlNode *Node) { // NOLINT
   for (xmlNode *H = Node->children; H; H = H->next) {
     if (H->type == XML_ELEMENT_NODE) {
       if (!xmlStrcmp(H->name, XmlConstants::CONSTRAINT)) {
-        string Cnt = std::string(
-            reinterpret_cast<char *>(std::unique_ptr<xmlChar, void (*)(void *)>(
-                                         xmlNodeGetContent(H), xmlFree)
-                                         .get()));
+        string Cnt = std::string(reinterpret_cast<char *>(
+            UniqueXmlChar(xmlNodeGetContent(H), xmlFree).get()));
```
```suggestion
        std::string Cnt{reinterpret_cast<char *>(
            UniqueXmlChar(xmlNodeGetContent(H), xmlFree).get())};
```

vulder (commented):
Just found a few nits. Otherwise, this looks good now. But let's wait for the Modification/Transaction interface before we merge this.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -1,7 +1,11 @@
 #include "vara/Feature/FeatureModelParser.h"
 
+#include "llvm/ADT/StringExtras.h"
+#include "llvm/Support/FormatVariadic.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include "SxfmConstants.h"
+
```
```suggestion
```

vulder (approved):
Looks good from my side, I added Lauritz to double check parsing things.

ChristianKaltenecker:
> Looks good from my side, I added Lauritz to double check parsing things.

Yes, good idea! Thanks! :blush: :+1:

lib/Feature/FeatureModelParser.cpp
```diff
@@ -286,10 +275,364 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+                                std::get<2>(OrGroup->second),
+                                std::get<0>(OrGroup->second));
+        OrGroupMapping.erase(CurrentIndentationLevel);
+      }
+
+      // Remember the new or group parent if there is one
+      if (Cardinalities.has_value()) {
+        Relationship::RelationshipKind GroupKind =
+            Relationship::RelationshipKind::RK_ALTERNATIVE;
+        if (std::get<1>(Cardinalities.value()) == UINT_MAX) {
```
s9latimm:
```
warning: comparison of integers of different signs: '__tuple_element_t<1UL, tuple<int, int>>' (aka 'int') and 'unsigned int' [-Wsign-compare]
        if (std::get<1>(Cardinalities.value()) == UINT_MAX) {
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -286,10 +275,364 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+          .substr(Pos + 1, CardinalityString.find(']', Pos + 1) - Pos - 1)
+          .str());
+
+  if (!MinCardinality.has_value() || !MaxCardinality.has_value()) {
+    llvm::errs() << "No parsable cardinality!\n";
+    return std::optional<std::tuple<int, int>>();
+  }
+
+  if (MinCardinality.value() != 1 ||
+      (MaxCardinality.value() != 1 && MaxCardinality.value() != UINT_MAX)) {
```
s9latimm:
```
warning: comparison of integers of different signs: 'int' and 'unsigned int' [-Wsign-compare]
      (MaxCardinality.value() != 1 && MaxCardinality.value() != UINT_MAX)) {
                                      ~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -286,10 +275,364 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
+  }
+  return true;
+}
+
+bool FeatureModelSxfmParser::parseFeatureTree(xmlChar *FeatureTree) {
+  // Split the lines of the feature tree by new lines
+  {
+    std::stringstream Ss(reinterpret_cast<const char *>(FeatureTree));
+    std::string To;
+    string Name;
```
s9latimm:
```suggestion
    std::string To;
    std::string Name;
```
It looks a bit strange, using one time `std::string` and the other time just `string`, but that is also pretty inconsistent throughout the whole file (including my code^^).

s9latimm (dismissed):
Somehow I get memory leaks through `xmlBufResize` and `xmlBufCreateSize` (which is new compared to `vara-dev`, not sure of the cause yet).

I am not sure, if using `UINT_MAX` as wildcard is really the best approach, but it works -- except for the signed comparisons commented below.

Overall the parser implementation LGTM and should work as intended.

include/vara/Feature/FeatureModelParser.h
Kallistos: I think we should either introduce a subclass of the FeatureModelParser that puts all commonalities of XML (like) Parser together or introduce these short cut in both (XML and SXFM Parser) as these types seem to be very specific to Parser for XML (-like) FIles, e.g. if we add a feature model parser for a CNF/SMT format, we would not need these short cuts, right?

include/vara/Feature/FeatureModelParser.h
```diff
@@ -61,8 +63,91 @@ class FeatureModelXmlParser : public FeatureModelParser {
...
+  ///
+  /// \returns true iff parsing and processing the constraints was successful
+  static bool parseConstraints(xmlChar *Constraints);
+
+  /// This method extracts the cardinality from the given line.
+  /// The cardinality is wrapped in square brackets (e.g., [1,1])
+  ///
+  /// \param StringToExtractFrom the string to extract the cardinality from
+  ///
+  /// \returns returns the cardinality of the given string and is empty if the
```
Kallistos: remove one returns

lib/Feature/FeatureModel.cpp
```diff
@@ -173,7 +173,7 @@ void FeatureModelBuilder::detectXMLAlternatives() {
     std::vector<std::string> Frontier(Children[FeatureName].begin(),
                                       Children[FeatureName].end());
     while (!Frontier.empty()) {
-      const auto &FName = Frontier.back();
+      const std::string FName{Frontier.back()};
```
Kallistos, line 176: why not `const std::string &` ?

lib/Feature/FeatureModelParser.cpp
```diff
@@ -286,10 +275,364 @@ std::unique_ptr<xmlDoc, void (*)(xmlDocPtr)> FeatureModelXmlParser::parseDoc() {
...
 bool FeatureModelXmlParser::verifyFeatureModel() { return parseDoc().get(); }
 
+//===----------------------------------------------------------------------===//
+//                        FeatureModelSxfmParser Class
+//===----------------------------------------------------------------------===//
+
+std::unique_ptr<FeatureModel> FeatureModelSxfmParser::buildFeatureModel() {
+  UniqueXmlDoc Doc = parseDoc();
+  if (!Doc) {
+    return nullptr;
```
Kallistos, line 291: Why do you return a `nullptr` here?
maybe throw an error or call a method that terminates the parser?
Or is this expected valid behaviour, that the Doc was empty?

lib/Feature/FeatureModelParser.cpp
Kallistos: UINT_MAX is in this case '-1' if that is what you wanted you can replace UINT_MAX with -1, or cast UINT_MAX explicitly to int.

ChristianKaltenecker: You are right. Using the cast obfuscates the real intention, so I will replace it with '-1'.

lib/Feature/FeatureModelParser.cpp
s9latimm: `buildFeatureModel` is actually the entry point for the parser, so returning here terminates the parser as well.

I returned `nullptr` for the XML part at this point too, as it is easy to test afterwards if parsing succeeded and somewhat recoverable, but I think an error message would not harm.

On the other hand, an empty input will always be rejected as it would be invalidated through the DTD.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: I have changed that. Now it is consistent; at least for the feature model parser.

lib/Feature/FeatureModel.cpp
ChristianKaltenecker: See https://github.com/se-sic/vara-feature/pull/29#discussion_r545079723

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: 1. The behavior is consistent to what the XML parser does. 
2. IF line 291 is reached, then errors are already printed out on the console. Error handling is done in the `parseDoc()` method that is called in line 289. I do not think that we need further error handling. 

But I am open for suggestions :blush:

ChristianKaltenecker:
> Somehow I get memory leaks through `xmlBufResize` and `xmlBufCreateSize` (which is new compared to `vara-dev`, not sure of the cause yet).

THIS is strange. We should definitely solve this before merging into vara-dev.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: I have introduced a constant 'WILDCARD' to make it clear that the '-1' is used as a placeholder for wildcards in the cardinalities.

s9latimm (approved):

Kallistos (approved):

PR #32, opened 2021-01-01
Introduces FeatureModelTransaction API

vulder:


include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -0,0 +1,290 @@
...
+void addFeature(FeatureModel *FM, std::unique_ptr<Feature> NewFeature,
+                Feature *Parent = nullptr);
+
+//===----------------------------------------------------------------------===//
+//                    Transaction Implementation Details
+//===----------------------------------------------------------------------===//
+
+namespace detail {
+
+struct CopyTransactionMode {};
```
s9latimm:
```suggestion
class CopyTransactionMode {};
```
`CopyTransactionMode` defined as a struct here but previously declared as a class.

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -0,0 +1,290 @@
...
+struct ModifyTransactionMode {};
```
s9latimm:
```suggestion
class ModifyTransactionMode {};
```
`ModifyTransactionMode` defined as a struct here but previously declared as a class.

lib/Feature/FeatureModel.cpp
```diff
@@ -14,35 +14,32 @@ void FeatureModel::dump() const {
...
+}
+
+void FeatureModel::removeFeature(Feature &F) {
+  if (&F == Root) {
+    Root = nullptr;
   }
-  OrderedFeatures.insert(Value);
-  return true;
+  OrderedFeatures.remove(&F);
+  Features.erase(F.getName());
```
ChristianKaltenecker, line 34: Should we also ensure that the child feature no longer use this feature as parent? Or is this already the case?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -0,0 +1,290 @@
...
+  FeatureModelTransaction(FeatureModel &FM) : TransactionBaseTy(FM) {}
+};
+
+using FeatureModelCopyTransaction =
+    FeatureModelTransaction<detail::CopyTransactionMode>;
+using FeatureModelModifyTransaction =
+    FeatureModelTransaction<detail::ModifyTransactionMode>;
+
+//===----------------------------------------------------------------------===//
+//                            Modifiction Helpers
```
ChristianKaltenecker:
```suggestion
//                            Modification Helpers
```

lib/Feature/FeatureModel.cpp
vulder: Nope, the FM just offers private modifier methods that exactly change what requested, however, these are all private and only accessible to Modifications (after we refactored the builder). The Modification itself then guarantees and is tested to ensure that (e.g., AddFeature) does the correct thing, setting parent/child relation ships etc..

Furthermore, on top of that we have the transactions, which group to together a sequence of modifications and guarantee that these together produce a valid feature model.

s9latimm (approved):

ChristianKaltenecker (approved):

lib/Feature/FeatureModelParser.cpp
```diff
@@ -98,6 +97,9 @@ bool FeatureModelXmlParser::parseConfigurationOption(xmlNode *Node,
       }
     }
   }
+  if (Name == "root") {
```
s9latimm:
```suggestion

  // XML has those names specified as root nodes
  if (Name == "root" || Name == "base") {
    FMB.setRootName(Name);
```

padupr (approved):

PR #33, opened 2021-01-16
Add Feature Model Cloning

s9latimm:
- Adds deep cloning for `FeatureModel` and related data structures (`Constraint`, `FeatureSourceRange`). 
- Reworks `getParentFeature` replacing `getParent` with a template in `FeatureTreeNode`.

### DIFF
--- a/include/vara/Feature/Constraint.h
+++ b/include/vara/Feature/Constraint.h
@@ -68,6 +68,8 @@ public:
     return R;
   }
 
+  [[nodiscard]] virtual std::unique_ptr<Constraint> clone() = 0;
+
   [[nodiscard]] virtual std::string toString() const = 0;
 
   [[nodiscard]] virtual std::string toHTML() const { return toString(); }
@@ -113,6 +115,12 @@ public:
       : BinaryConstraint(ConstraintKind::CK_OR, std::move(LeftOperand),
                          std::move(RightOperand)) {}
 
+  std::unique_ptr<Constraint> clone() override {
+    return std::make_unique<OrConstraint>(
+        std::move(this->getLeftOperand()->clone()),
+        std::move(this->getRightOperand()->clone()));
+  }
+
   [[nodiscard]] std::string toString() const override {
     return llvm::formatv("({0} | {1})", LeftOperand->toString(),
                          RightOperand->toString());
@@ -130,6 +138,12 @@ public:
       : BinaryConstraint(ConstraintKind::CK_XOR, std::move(LeftOperand),
                          std::move(RightOperand)) {}
 
+  std::unique_ptr<Constraint> clone() override {
+    return std::make_unique<XorConstraint>(
+        std::move(this->getLeftOperand()->clone()),
+        std::move(this->getRightOperand()->clone()));
+  }
+
   [[nodiscard]] std::string toString() const override {
     return llvm::formatv("({0} ^ {1})", LeftOperand->toString(),
                          RightOperand->toString());
@@ -146,6 +160,12 @@ public:
                 std::unique_ptr<Constraint> RightOperand)
       : BinaryConstraint(ConstraintKind::CK_AND, std::move(LeftOperand),
                          std::move(RightOperand)) {}
+
+  std::unique_ptr<Constraint> clone() override {
+    return std::make_unique<AndConstraint>(
+        std::move(this->getLeftOperand()->clone()),
+        std::move(this->getRightOperand()->clone()));
+  }
 
   [[nodiscard]] std::string toString() const override {
     return llvm::formatv("({0} & {1})", LeftOperand->toString(),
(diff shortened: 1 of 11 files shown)


include/vara/Feature/Constraint.h
```diff
@@ -113,6 +115,12 @@ class OrConstraint : public BinaryConstraint, public BooleanConstraint {
       : BinaryConstraint(ConstraintKind::CK_OR, std::move(LeftOperand),
                          std::move(RightOperand)) {}
 
+  std::unique_ptr<Constraint> clone() override {
+    return std::make_unique<OrConstraint>(
+        std::move(this->getLeftOperand()->clone()),
+        std::move(this->getRightOperand()->clone()));
```
```suggestion
        this->getLeftOperand()->clone(),
        this->getRightOperand()->clone());
```
The returned value is already an rvalue

include/vara/Feature/Constraint.h
```diff
@@ -130,6 +138,12 @@ class XorConstraint : public BinaryConstraint, public BooleanConstraint {
       : BinaryConstraint(ConstraintKind::CK_XOR, std::move(LeftOperand),
                          std::move(RightOperand)) {}
 
+  std::unique_ptr<Constraint> clone() override {
+    return std::make_unique<XorConstraint>(
+        std::move(this->getLeftOperand()->clone()),
+        std::move(this->getRightOperand()->clone()));
```
```suggestion
        this->getLeftOperand()->clone(),
        this->getRightOperand()->clone());
```

include/vara/Feature/Constraint.h
```diff
@@ -147,6 +161,12 @@ class AndConstraint : public BinaryConstraint, public BooleanConstraint {
       : BinaryConstraint(ConstraintKind::CK_AND, std::move(LeftOperand),
                          std::move(RightOperand)) {}
 
+  std::unique_ptr<Constraint> clone() override {
+    return std::make_unique<AndConstraint>(
```
vulder: etc...

include/vara/Feature/Constraint.h
```diff
@@ -481,10 +587,14 @@ class Feature;
 class PrimaryFeatureConstraint : public PrimaryConstraint {
 public:
   PrimaryFeatureConstraint(std::variant<Feature *, std::unique_ptr<Feature>> FV)
-      : PrimaryConstraint(ConstraintKind::CK_FEATURE), FV(std::move(FV))  {}
+      : PrimaryConstraint(ConstraintKind::CK_FEATURE), FV(std::move(FV)) {}
 
   [[nodiscard]] Feature *getFeature() const;
 
+  std::unique_ptr<Constraint> clone() override {
+    return std::make_unique<PrimaryFeatureConstraint>(this->getFeature());
```
vulder: Is this feature ptr later replaced?

include/vara/Feature/Feature.h
```diff
@@ -50,18 +50,6 @@ class Feature : public FeatureTreeNode {
 
   [[nodiscard]] bool isOptional() const { return Opt; }
 
-  /// Search parent feature in tree structure -- this may not exist or nullptr
-  /// if node is already root.
-  [[nodiscard]] Feature *getParentFeature() const {
```
vulder: Why did you move this to the `TreeNodeClass`? As an abstraction this makes more sense on a Feature, with the move the TreeNode is now hard Tied to the Feature.

include/vara/Feature/Feature.h
```diff
@@ -225,12 +213,13 @@ class BinaryFeature : public Feature {
   BinaryFeature(
       string Name, bool Opt = false,
       std::vector<FeatureSourceRange> Loc = std::vector<FeatureSourceRange>(),
-      Feature *Parent = nullptr, const NodeSetType &Children = {})
+      FeatureTreeNode *Parent = nullptr, const NodeSetType &Children = {})
```
vulder: A gain, more abstraction breaks. I don't think we should rework this.

vulder (changes requested):
Hmm, I think we should not break the Feature/FeatureTreeNode abstraction. Why did you mix these two to implement clone?

include/vara/Feature/Feature.h
s9latimm: `Relationship` now needs a look up to a parent feature too. I reworked this as a template in `FeatureTreeNode`.

include/vara/Feature/Constraint.h
s9latimm: I reworked this part, so that this copy only holds a new shallow feature (name only), which later needs replacement (the `FeatureModelBuilder` used cloning a `FeatureModel` already does exactly that).

Also, added comment should explain the intended behavior.

include/vara/Feature/Feature.h
vulder: I fixed the Relationship with local cast handling, we don't really need the template.

unittests/Feature/FeatureModel.cpp
```diff
@@ -13,6 +13,79 @@ TEST(FeatureModel, build) {
...
+  auto FM = B.buildFeatureModel();
+  assert(FM);
+
+  auto Clone = FM->clone();
+  assert(Clone);
+  FM.reset();
+
+  // NOLINTNEXTLINE
+  ASSERT_EXIT(EXPECT_TRUE(FM->getFeature("a")->isRoot()),
+              testing::KilledBySignal(SIGSEGV), ".*");
```
vulder, line 44: Why do you need the EXPECT_TRUE inside the EXIT? Should the segfault not also occur without it?

vulder (approved):

PR #34, opened 2021-01-28
AddFeature Modification

s9latimm:


include/vara/Feature/FeatureModel.h
```diff
@@ -154,28 +153,29 @@ class FeatureModelBuilder : private FeatureModel {
...
+      return nullptr;
+    }
+    return Features[FeatureName].get();
   }
 
-  bool addFeature(Feature &F);
-
-  FeatureModelBuilder *addParent(const std::string &FeatureName,
-                                 const std::string &ParentName) {
+  FeatureModelBuilder *addEdge(const std::string &ParentName,
```
vulder, line 177: Why did you rename this? `addParent` seemed fine to me

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -135,14 +135,13 @@ class FeatureModelModification {
 
 protected:
   /// \brief Set the parrent of a Feature.
-  static void setParent(Feature &F, Feature *Parent) {
-    F.setParent(Parent);
-  }
+  static void setParent(Feature &F, Feature *Parent) { F.setParent(Parent); }
 
-  static void addChild(Feature &F, Feature &Child) {
+  static void addChild(Feature &F, Feature *Child) {
```
```suggestion
  static void addChild(Feature &F, Feature &Child) {
```
There is no go way to handle nullptrs here, so why should `Child` be a ptr?

lib/Feature/FeatureModel.cpp
```diff
@@ -14,28 +14,15 @@ void FeatureModel::dump() const {
...
 
-bool FeatureModel::addFeature(std::unique_ptr<Feature> NewFeature) {
-  // TODO(s9latimm): check consistency
+Feature *FeatureModel::addFeature(std::unique_ptr<Feature> NewFeature) {
   auto FeatureName = std::string(NewFeature->getName());
   if (!Features.try_emplace(FeatureName, std::move(NewFeature)).second) {
-    return false;
+    return nullptr;
   }
   auto *InsertedFeature = Features[FeatureName].get();
```
vulder: If I remember correctly `try_emplace` returns and iterator, bool pair.
You create a local var with this and then use it to make the check and you don't have to lookup the feature.
```
auto PosInsertedFeature = Features.try_emplace(...)
if (PosInsertedFeature.second) {
  return nullptr;
}
...
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -30,7 +30,9 @@ bool FeatureModelXmlParser::parseConfigurationOption(xmlNode *Node,
       } else if (!xmlStrcmp(Head->name, XmlConstants::OPTIONAL)) {
         Opt = Cnt == "True";
       } else if (!xmlStrcmp(Head->name, XmlConstants::PARENT)) {
-        FMB.addParent(Name, Cnt);
+        // TODO(s9latimm) fix me -- breaks if Name not set / parsed yet
```
vulder, line 33: What should we do in this case?

unittests/Feature/FeatureModel.cpp
```diff
@@ -48,14 +50,9 @@ TEST(FeatureModel, cloneRoot) {
 
 TEST(FeatureModel, cloneRelationship) {
   FeatureModelBuilder B;
-  BinaryFeature AA("aa");
-  BinaryFeature AB("ab");
-  auto CS = Feature::NodeSetType();
-  CS.insert(&AA);
-  CS.insert(&AB);
-  B.addParent("aa", "root")->addFeature(AA);
-  B.addParent("ab", "root")->addFeature(AB);
```
vulder, line 57: I guess the parent relation is not strictly needed in this test or?

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -16,42 +16,58 @@ class FeatureModelModificationTest : public ::testing::Test,
...
   void SetUp() override {
     FeatureModelBuilder B;
-    B.addParent("a", "root")->makeFeature<BinaryFeature>("a", true);
+    B.makeFeature<BinaryFeature>("a");
     FM = B.buildFeatureModel();
   }
 
   // Dummy method to fullfill the FeatureModelModification interface
-  void exec(FeatureModel &FM) override {};
+  void exec(FeatureModel &FM) override{};
```
vulder, line 24:
```suggestion
  void exec(FeatureModel &FM) override {};
```
What version of clang-format are you using?

vulder (changes requested):
Did you adapt the FeatureModelBuilder to also use the modifications? Where there problems?

lib/Feature/FeatureModelParser.cpp
s9latimm: We could search `name`  beforehand with an initial pass over all `Node->children`, unfortunately that way we have to look at each node twice (or at least until we find `name`). I will add a separate PR from `vara-dev` with the fix, so that we can resolve this quickly.

include/vara/Feature/FeatureModel.h
s9latimm: I wanted to clarify, that calling this implies, that the children relation between these nodes will be updated too (while building the tree) -- in contrast to the `FeatureModelModification` where we implemented `setParent` so it literally changes only that.

unittests/Feature/FeatureModel.cpp
s9latimm: As we add those features without changing root, those edges are already implied (so is the existence of a root node, which it neither of those two).

unittests/Feature/FeatureModelTransaction.cpp
s9latimm: I updated today to `13.0.0` (latest from `llvm-project/main`) but above change is still persistent (with even more similar diffs according to changed WS rules).

include/vara/Feature/FeatureModelTransaction.h
s9latimm: As we talked about it last week already -- there is no reason, as any `nullptr` should be handled beforehand -- will revert that and update references accordingly.

unittests/Feature/FeatureModelTransaction.cpp
vulder: hmm, kk

PR #37, opened 2021-02-06
Add Reviewdog Format Suggestions

s9latimm:
Update workflows to comment automated formatting suggestions on pull requests.

- [x] Add workflow with `reviewdog` integration to suggest `clang-format` changes
  - [x] Move meta data into strategy
- [x] Update `build` workflow for consistency
  - [x] Force `--no-install-recommends`
  - [x] Update README with new badge and link to actual actions list
  - [x] Update required check for protected branches (@vulder)
- [x] Test reformatting with example
  - [x] Test different scenarios
  - [x] Remove example

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -6,7 +6,22 @@ on:
 
 jobs:
   format:
-    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        tool: [ clang-format ]
+        include:
+          - tool: clang-format
+            os: ubuntu-20.04
+            install: |
+              sudo apt-key adv --fetch-keys https://apt.llvm.org/llvm-snapshot.gpg.key
+              sudo add-apt-repository -y 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main'
+              sudo apt-get update
+              sudo apt-get -y install --no-install-recommends clang-format-11
+            regex: \.(h|c|hpp|cpp)$
+            command: clang-format-11 -style=llvm -i
+
+    runs-on: ${{ matrix.os }}
     continue-on-error: false
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,5 +1,5 @@
 # VaRA feature library
-[![Build Status](https://github.com/se-sic/vara-feature/workflows/VaRA%20feature%20library%20CI/badge.svg?branch=vara-dev)](https://github.com/se-sic/vara-feature) [![Documentation Status](https://readthedocs.org/projects/vara/badge/?version=vara-dev)](https://vara.readthedocs.io/en/vara-dev/?badge=vara-dev)
+[![Build Status](https://github.com/se-sic/vara-feature/workflows/VaRA%20Feature%20Library%20Build/badge.svg?branch=vara-dev)](https://github.com/se-sic/vara-feature/actions?query=branch%3Avara-dev) [![Documentation Status](https://readthedocs.org/projects/vara/badge/?version=vara-dev)](https://vara.readthedocs.io/en/vara-dev/?badge=vara-dev)
```
vulder, line 2: I don't think we need a batch for formatting.

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Fetch Base Branch
+        shell: bash
+        run: |
+          git fetch -n -p --depth=1 --no-recurse-submodules origin +refs/heads/${{ github.base_ref }}:refs/remotes/origin/${{ github.base_ref }}
+
+      - name: Run Clang-Format
+        shell: bash
+        run: |
+          FILES=$(git diff --name-only --diff-filter=AM origin/${{ github.base_ref }} HEAD | egrep '\.(c|h|hpp|cpp)$' || true)
+          [ -z "$FILES" ] || (clang-format-11 -i $FILES && git diff --diff-filter=M $FILES)
```
vulder: For maintainability it would be better to have the clang-format-11 command in a variable defined above so it's easier to see where we need to update things if necessary

vulder (changes requested):
Look's good so far.

We should remove the test file before merging.

README.md
s9latimm: This is only an update of the existing `build` badge with the changed workflow name and pointing to the related actions page.

Additionally the `format`-job currently always succeeds, so a badge would not make much sense anyway.

.github/workflows/reviewdog.yml
s9latimm: How about moving `clang-format-11 -i` and the file extension regex into an strategy, so that we are independent of types?

README.md
vulder: I see :+1:

.github/workflows/reviewdog.yml
s9latimm: I changed the job structure, so that the steps are independent of the actual tool, which itself is defined in a strategy.
This allows us to use arbitrary reformatting tools for other languages as well (for example python) with little modification.

vulder: very nice

vulder (approved):
Nice work :+1:

PR #39, opened 2021-02-09
Add Modifications and RootFeature

s9latimm:
Add modifications to test broken FMs, refactor root nodes and clean up public interface.

- [x] Add removal of edges, children and features as modifications
  - [x] Enable creation of broken FM through inheritance
- [x] Remove `Feature::isRoot` to prevent misconception
  - [x] Add `RootFeature` class
  - [x] Refactor builder
  - [x] Refactor comparisons to root with `llvm::isa`
- [x] Clean up unused / deprecated constructors
- [x] Update python bindings
  - [x] Add `RootFeature`
  - [x] Add `feature_model.get_feature`
  - [x] Clean up modifications / deprecated methods
- [x] add `venv/` to gitignore

### DIFF
--- a/bindings/python/vara-feature/pybind_Feature.cpp
+++ b/bindings/python/vara-feature/pybind_Feature.cpp
@@ -99,9 +99,17 @@ void init_feature_module_numeric_feature(py::module &M) {
           R"pbdoc(Returns the string representation of a NumericFeature.)pbdoc");
 }
 
+void init_feature_module_root_feature(py::module &M) {
+  py::class_<vf::RootFeature, vf::Feature>(M, "RootFeature")
+      .def(py::init<std::string>())
+      .def("to_string", &vf::RootFeature::toString,
+           R"pbdoc(Returns the string representation of a RootFeature.)pbdoc");
+}
+
 void init_feature_module(py::module &M) {
   init_feature_module_feature_tree_node(M);
   init_feature_module_feature(M);
   init_feature_module_binary_feature(M);
   init_feature_module_numeric_feature(M);
+  init_feature_module_root_feature(M);
 }
--- a/bindings/python/vara-feature/pybind_Feature.cpp
+++ b/bindings/python/vara-feature/pybind_Feature.cpp
@@ -31,10 +31,6 @@ void init_feature_module_feature_tree_node(py::module &M) {
 
       //===----------------------------------------------------------------===//
       // Parent
-      .def(
-          "has_parent",
-          [](vf::FeatureTreeNode &F) { return F.getParent() != nullptr; },
-          R"pbdoc(`True` if this is not the root of a `FeatureModel`.)pbdoc")
       .def("parent", &vf::FeatureTreeNode::getParent,
            py::return_value_policy::reference, R"pbdoc(Parent feature)pbdoc")
       .def("is_parent", &vf::FeatureTreeNode::hasEdgeFrom,
@@ -46,6 +42,10 @@ void init_feature_module_feature(py::module &M) {
       .def("__hash__", &vf::Feature::hash)
       .def_property_readonly("name", &vf::Feature::getName,
                              R"pbdoc(The name of the feature.)pbdoc")
+      .def(
+          "is_root",
+          [](vf::Feature &F) { return llvm::isa<vf::RootFeature>(F); },
+          R"pbdoc(`True` if this is the root of a `FeatureModel`.)pbdoc")
       .def("is_optional", &vf::Feature::isOptional,
            R"pbdoc(`True` if the feature is optional.)pbdoc")
       .def_property_readonly(
--- a/bindings/python/vara-feature/pybind_Feature.cpp
+++ b/bindings/python/vara-feature/pybind_Feature.cpp
@@ -26,13 +26,14 @@ void init_feature_module_feature_tree_node(py::module &M) {
             return py::make_iterator(F.begin(), F.end());
           },
           py::keep_alive<0, 1>())
-      .def("has_child", &vf::FeatureTreeNode::hasEdgeTo,
+      .def("is_child", &vf::FeatureTreeNode::hasEdgeTo,
            R"pbdoc(Checks if a node is a child of this one.)pbdoc")
 
       //===----------------------------------------------------------------===//
       // Parent
       .def(
-          "has_parent", [](vf::FeatureTreeNode &F) { return F.getParent(); },
+          "has_parent",
+          [](vf::FeatureTreeNode &F) { return F.getParent() != nullptr; },
           R"pbdoc(`True` if this is not the root of a `FeatureModel`.)pbdoc")
       .def("parent", &vf::FeatureTreeNode::getParent,
            py::return_value_policy::reference, R"pbdoc(Parent feature)pbdoc")
--- a/bindings/python/vara-feature/pybind_Feature.cpp
+++ b/bindings/python/vara-feature/pybind_Feature.cpp
@@ -44,7 +44,7 @@ void init_feature_module_feature(py::module &M) {
                              R"pbdoc(The name of the feature.)pbdoc")
       .def(
           "is_root",
-          [](vf::Feature &F) { return llvm::isa<vf::RootFeature>(F); },
+          [](const vf::Feature &F) { return llvm::isa<vf::RootFeature>(F); },
           R"pbdoc(`True` if this is the root of a `FeatureModel`.)pbdoc")
       .def("is_optional", &vf::Feature::isOptional,
            R"pbdoc(`True` if the feature is optional.)pbdoc")
--- a/bindings/python/vara-feature/pybind_Feature.cpp
+++ b/bindings/python/vara-feature/pybind_Feature.cpp
@@ -12,9 +12,6 @@ namespace py = pybind11;
 
 void init_feature_module_feature_tree_node(py::module &M) {
   py::class_<vf::FeatureTreeNode>(M, "FeatureTreeNode")
-      .def("is_root", &vf::FeatureTreeNode::isRoot,
-           R"pbdoc(`True` if this is the root of a `FeatureModel`.)pbdoc")
-
       //===----------------------------------------------------------------===//
       // Children
       .def(
@@ -29,11 +26,14 @@ void init_feature_module_feature_tree_node(py::module &M) {
             return py::make_iterator(F.begin(), F.end());
           },
           py::keep_alive<0, 1>())
-      .def("is_child", &vf::FeatureTreeNode::hasEdgeTo,
+      .def("has_child", &vf::FeatureTreeNode::hasEdgeTo,
            R"pbdoc(Checks if a node is a child of this one.)pbdoc")
 
       //===----------------------------------------------------------------===//
       // Parent
+      .def(
+          "has_parent", [](vf::FeatureTreeNode &F) { return F.getParent(); },
+          R"pbdoc(`True` if this is not the root of a `FeatureModel`.)pbdoc")
       .def("parent", &vf::FeatureTreeNode::getParent,
            py::return_value_policy::reference, R"pbdoc(Parent feature)pbdoc")
       .def("is_parent", &vf::FeatureTreeNode::hasEdgeFrom,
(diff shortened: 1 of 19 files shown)


bindings/python/vara-feature/pybind_Feature.cpp
```diff
@@ -80,13 +81,6 @@ void init_feature_module_binary_feature(py::module &M) {
       .def(py::init<std::string, bool>())
       .def(py::init<std::string, bool,
                     std::vector<vara::feature::FeatureSourceRange>>())
-      .def(py::init<std::string, bool,
-                    std::vector<vara::feature::FeatureSourceRange>,
-                    vara::feature::Feature *>())
-      .def(py::init<std::string, bool,
-                    std::vector<vara::feature::FeatureSourceRange>,
-                    vara::feature::Feature *,
-                    std::vector<vara::feature::FeatureTreeNode *>>())
```
s9latimm, line 89: @padupr do you need these constructors?

bindings/python/vara-feature/pybind_Feature.cpp
```diff
@@ -45,6 +42,10 @@ void init_feature_module_feature(py::module &M) {
       .def("__hash__", &vf::Feature::hash)
       .def_property_readonly("name", &vf::Feature::getName,
                              R"pbdoc(The name of the feature.)pbdoc")
+      .def(
+          "is_root",
+          [](vf::Feature &F) { return llvm::isa<vf::RootFeature>(F); },
```
vulder:
```suggestion
          [](const vf::Feature &F) { return llvm::isa<vf::RootFeature>(F); },
```
I, think this should be const, otherwise, we could get a problem when we call python functions on an underlying const object.

bindings/python/vara-feature/pybind_FeatureModel.cpp
```diff
@@ -22,6 +22,12 @@ void init_feature_model_module(py::module &M) {
       .def("get_root", &vf::FeatureModel::getRoot,
            py::return_value_policy::reference,
            R"pbdoc(Returns the root Feature.)pbdoc")
+      .def(
+          "get_feature",
+          [](vf::FeatureModel &FM, const std::string &Name) {
```
vulder, line 27: Do we get a ref here from the python side? Not sure, normally, if we would call a member function on an C++ object this would be a pointer.

include/vara/Feature/Feature.h
```diff
@@ -278,6 +257,21 @@ class NumericFeature : public Feature {
...
   ValuesVariantType Values;
 };
+
+//===----------------------------------------------------------------------===//
+//                          RootFeature Class
+//===----------------------------------------------------------------------===//
+
+class RootFeature : public Feature {
+public:
+  RootFeature(string Name = "root")
```
```suggestion
  explicit RootFeature(string Name = "root")
```
Ctors with only one parameter should be explicit to prevent unwanted implicit conversions.

include/vara/Feature/FeatureModel.h
```diff
@@ -438,16 +446,17 @@ template <typename... Rules> class FeatureModelConsistencyChecker {
...
   }
 };
 
 struct CheckFeatureParentChildRelationShip {
   static bool check(FeatureModel &FM) {
     return std::all_of(FM.begin(), FM.end(), [](Feature *F) {
-      return F->isRoot() ||
-             // Every parent of a Feature needs to have the Feature as a child.
+      return llvm::isa<RootFeature>(F) ||
+             // Every parent of a Feature needs to have that as a child.
```
```suggestion
             // Every parent of a Feature needs to have it as a child.
```

include/vara/Feature/FeatureModel.h
```diff
@@ -438,16 +446,17 @@ template <typename... Rules> class FeatureModelConsistencyChecker {
 struct EveryFeatureRequiresParent {
   static bool check(FeatureModel &FM) {
     return std::all_of(FM.begin(), FM.end(), [](Feature *F) {
-      return F->isRoot() || (F->getParent() != nullptr);
+      return llvm::isa<RootFeature>(F) ||
+             (F->getParent() && F->getParentFeature());
```
vulder: Why do we need the extra `getParentFeature` check?

include/vara/Feature/FeatureModel.h
```diff
@@ -164,15 +166,15 @@ class FeatureModelBuilder : private FeatureModel {
   template <typename FeatureTy, typename... Args,
             typename = typename std::enable_if_t<
                 std::is_base_of_v<Feature, FeatureTy>, int>>
-  Feature *makeFeature(const std::string &FeatureName, Args... FurtherArgs) {
+  FeatureTy *makeFeature(const std::string &FeatureName, Args... FurtherArgs) {
     if (!Features
              .try_emplace(FeatureName,
                           std::make_unique<FeatureTy>(
                               FeatureName, std::move(FurtherArgs)...))
```
```suggestion
  FeatureTy *makeFeature(const std::string &FeatureName, Args &&...FurtherArgs) {
    if (!Features
             .try_emplace(FeatureName,
                          std::make_unique<FeatureTy>(
                              FeatureName, std::forward<Args>(FurtherArgs)...))
```

include/vara/Feature/FeatureModel.h
```diff
@@ -164,15 +166,15 @@ class FeatureModelBuilder : private FeatureModel {
   template <typename FeatureTy, typename... Args,
             typename = typename std::enable_if_t<
                 std::is_base_of_v<Feature, FeatureTy>, int>>
```
```suggestion
  template <typename FeatureTy, typename... Args,
            std::enable_if_t<std::is_base_of_v<Feature, FeatureTy>, int> = 0>
```

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -22,7 +22,8 @@ class OrderedFeatureVector {
   OrderedFeatureVector(std::initializer_list<Feature *> Init) { insert(Init); }
   template <class FeatureIterTy>
   OrderedFeatureVector(FeatureIterTy Begin, FeatureIterTy End) {
```
vulder:
```suggestion
  OrderedFeatureVector(FeatureIterTy &&Begin, FeatureIterTy &&End) {
```

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -42,7 +49,8 @@ class OrderedFeatureVector {
 
   template <class FeatureIterTy>
   void insert(FeatureIterTy Begin, FeatureIterTy End) {
```
vulder:
```suggestion
  void insert(FeatureIterTy &&Begin, FeatureIterTy &&End) {
```

include/vara/Feature/FeatureModel.h
```diff
@@ -164,15 +166,15 @@ class FeatureModelBuilder : private FeatureModel {
...
-  Feature *makeFeature(const std::string &FeatureName, Args... FurtherArgs) {
+  FeatureTy *makeFeature(const std::string &FeatureName, Args... FurtherArgs) {
```
```suggestion
  FeatureTy *makeFeature(std::string FeatureName, Args... FurtherArgs) {
```

unittests/Feature/BinaryFeature.cpp
```diff
@@ -44,9 +42,7 @@ TEST(BinaryFeature, BinaryFeatureChildren) {
   EXPECT_EQ(
       std::distance(FM->getFeature("a")->begin(), FM->getFeature("a")->end()),
       1);
-  if (auto *F =
-          llvm::dyn_cast<vara::feature::Feature>(*FM->getFeature("a")->begin());
-      F) {
+  if (auto *F = llvm::dyn_cast<Feature>(*FM->getFeature("a")->begin()); F) {
```
vulder:
```suggestion
  if (auto *F = llvm::dyn_cast<Feature>(*FM->getFeature("a")->begin())) {
```
If you don't have a more complex check of need the variable defined in the else branch you can just use an normal if here without initializer.

unittests/Feature/FeatureModel.cpp
```diff
@@ -45,20 +34,19 @@ TEST(FeatureModel, cloneUnique) {
...
   FM.reset();
 
-  // The inner EXPECT_TRUE just handles the nodiscard of isRoot
+  // The inner EXPECT_TRUE just handles the nodiscard of getParent
   // NOLINTNEXTLINE
-  ASSERT_EXIT(EXPECT_TRUE(FM->getFeature("a")->isRoot()),
-              testing::KilledBySignal(SIGSEGV), ".*");
-  EXPECT_TRUE(Clone->getFeature("a")->isRoot());
+  EXPECT_DEATH(EXPECT_TRUE(FM->getFeature("a")->getParent()), ".*");
+  EXPECT_FALSE(Clone->getFeature("a")->getParent());
```
vulder: Shouldn't this become a `isa` check for RootNode instead of an empty parent check?

unittests/Feature/NumericFeature.cpp
```diff
@@ -60,9 +58,7 @@ TEST(NumericFeature, NumericFeatureChildren) {
   EXPECT_EQ(
       std::distance(FM->getFeature("a")->begin(), FM->getFeature("a")->end()),
       1);
-  if (auto *F =
-          llvm::dyn_cast<vara::feature::Feature>(*FM->getFeature("a")->begin());
-      F) {
+  if (auto *F = llvm::dyn_cast<Feature>(*FM->getFeature("a")->begin()); F) {
```
vulder:
```suggestion
  if (auto *F = llvm::dyn_cast<Feature>(*FM->getFeature("a")->begin())) {
```

unittests/Feature/RootFeature.cpp
```diff
@@ -0,0 +1,54 @@
...
+  assert(FM);
+
+  EXPECT_EQ(
+      std::distance(FM->getFeature("a")->begin(), FM->getFeature("a")->end()),
+      1);
+  if (auto *F = llvm::dyn_cast<Feature>(*FM->getFeature("a")->begin()); F) {
+    EXPECT_EQ("aa", F->getName());
+  } else {
+    FAIL();
+  }
```
vulder:
```suggestion
  Feature *FirstChild = llvm::dyn_cast<Feature>(*FM->getFeature("a")->begin());
  EXPECT_NE(FirstChild, nullptr);
  EXPECT_EQ("aa", F->getName());
```
This produces more precise error messages.

vulder (changes requested):

include/vara/Feature/FeatureModel.h
s9latimm: In this loop we only iterate over feature nodes, so relationships parents are not covered. If we assert, that each feature has an ancestor, this implies that also every relationship has a parent of some kind (assuming we prohibit directly nested relationships).

unittests/Feature/FeatureModel.cpp
s9latimm: Indeed.

unittests/Feature/RootFeature.cpp
s9latimm: I agree, but Clang-Tidy wants me to assure the pointer cannot be null in this block -- not even `ASSERT_` satisfies here.

include/vara/Feature/FeatureModel.h
s9latimm: Moving `FeatureName` during the creation breaks, as it is needed for the emplacement afterwards.

bindings/python/vara-feature/pybind_FeatureModel.cpp
s9latimm: So far we always used refs as arguments and looking at pybind11s documentation, they are doing so as well -- idk.
It seems like ptrs work too, at least they do not produce errors yet.

vulder: ok then let's stay with refs

include/vara/Feature/FeatureModel.h
vulder: But how does checking if the parent is a Feature help if we don't iterate over the other nodes?

unittests/Feature/RootFeature.cpp
vulder: Does a assert(FirstChild) help?

s9latimm: Yes, but `Release` will still fail.

vulder: ok, your right, then let's keep this.

include/vara/Feature/FeatureModel.h
s9latimm: If we have a feature in a relationship, the parent node will only be of that kind. So if we find an parent feature (which can't be the actual parent), this means, that the relationship node itself has to have a parent, which again should be a feature. So we implicitly check if every relationship has a parent as well.

tl;dr: If we check that features are connected to a parent feature, this may be transitive over a relationship -- so the relationship itself has to have a parent in the first place.

include/vara/Feature/FeatureModel.h
vulder: ok I remove the move

bindings/python/vara-feature/pybind_Feature.cpp
vulder: nope, je does not need them

padupr: I don't think so.

PR #27, opened 2020-12-08
XML extensions

padupr:
resolves se-passau/VaRA#697
resolves se-passau/VaRA#707
resolves se-passau/VaRA#708

PR #42, opened 2021-02-22
Fixes for last xml extension

padupr:
Completes #27 

- commit getter
- Location constructor with default category
- tests for python bindings


padupr:
There should be a setter for the commit value, should this be a transaction?

vulder (approved):

include/vara/Feature/FeatureModel.h
```diff
@@ -58,6 +58,8 @@ class FeatureModel {
 
   [[nodiscard]] llvm::StringRef getCommit() const { return Commit; }
 
+  void setCommit(llvm::StringRef Str) { Commit = Str; }
```
```suggestion
  void setCommit(std::string Str) { Commit = std::move(Str); }
```

bindings/python/vara-feature/pybind_FeatureModel.cpp
```diff
@@ -19,6 +19,13 @@ void init_feature_model_module(py::module &M) {
       .def_property_readonly(
           "path",
           [](const vf::FeatureModel &FM) { return FM.getPath().string(); })
+      .def_property(
+          "commit",
+          [](const vf::FeatureModel &FM) { return FM.getCommit().str(); },
+          [](vf::FeatureModel &FM, std::string &Commit) {
```
vulder: Use as string value and move it or a `const &` but value should work better here.

vulder (changes requested):

vulder (approved):

PR #43, opened 2021-02-25
Refactor builder with transactions

s9latimm:
Implement FMBuilder using transactions while adding required modifications.

- [x] Refactor `FeatureModelBuilder`
  - [x] move to own file
  - [x] `AddChild`
  - [x] `SetRoot`
  - [x] `SetPath`
  - [x] `SetCommit`
  - [x] `SetName`
  - [x] `AddConstraintToModel` (see se-passau/VaRA/issues/664)
  - [x] `AddRelationshipToModel` (see #44)
    - [x] tests 
- [x] Fix failing tests
- [x] Clang-Tidy

Related se-passau/VaRA/issues/744

Closes se-passau/VaRA/issues/635
Closes se-passau/VaRA/issues/669
Closes se-passau/VaRA/issues/692
Closes se-passau/VaRA/issues/724

### DIFF
--- a/include/vara/Feature/Constraint.h
+++ b/include/vara/Feature/Constraint.h
@@ -572,6 +572,10 @@ private:
 
 class Feature;
 
+namespace detail {
+class FeatureModelModification;
+} // namespace detail
+
 class PrimaryFeatureConstraint : public PrimaryConstraint {
 public:
   PrimaryFeatureConstraint(std::variant<Feature *, std::unique_ptr<Feature>> FV)
@@ -595,7 +599,7 @@ public:
   void accept(ConstraintVisitor &V) override;
 
 private:
-  friend FeatureModelBuilder;
+  friend detail::FeatureModelModification;
 
   std::variant<Feature *, std::unique_ptr<Feature>> FV;
 
--- a/include/vara/Feature/Constraint.h
+++ b/include/vara/Feature/Constraint.h
@@ -577,6 +577,8 @@ class FeatureModelModification;
 } // namespace detail
 
 class PrimaryFeatureConstraint : public PrimaryConstraint {
+  friend detail::FeatureModelModification;
+
 public:
   PrimaryFeatureConstraint(std::variant<Feature *, std::unique_ptr<Feature>> FV)
       : PrimaryConstraint(ConstraintKind::CK_FEATURE), FV(std::move(FV)) {}
@@ -599,8 +601,6 @@ public:
   void accept(ConstraintVisitor &V) override;
 
 private:
-  friend detail::FeatureModelModification;
-
   std::variant<Feature *, std::unique_ptr<Feature>> FV;
 
   void setFeature(Feature *F) { this->FV = F; }
--- a/include/vara/Feature/Feature.h
+++ b/include/vara/Feature/Feature.h
@@ -92,9 +92,7 @@ public:
   // Locations
   [[nodiscard]] bool hasLocations() const { return !Locations.empty(); }
 
-  void addLocation(FeatureSourceRange &Fsr) {
-    Locations.push_back(std::move(Fsr));
-  }
+  void addLocation(FeatureSourceRange Fsr) { Locations.push_back(Fsr); }
   std::vector<FeatureSourceRange>::iterator
   removeLocation(const FeatureSourceRange &Fsr) {
     return Locations.erase(std::find(Locations.begin(), Locations.end(), Fsr));
--- a/include/vara/Feature/Feature.h
+++ b/include/vara/Feature/Feature.h
@@ -265,7 +265,7 @@ private:
 
 class RootFeature : public Feature {
 public:
-  explicit RootFeature(string Name = "root")
+  explicit RootFeature(string Name)
       : Feature(FeatureKind::FK_ROOT, std::move(Name), false, {}) {}
 
   static bool classof(const Feature *F) {
--- a/include/vara/Feature/Feature.h
+++ b/include/vara/Feature/Feature.h
@@ -92,7 +92,10 @@ public:
   // Locations
   [[nodiscard]] bool hasLocations() const { return !Locations.empty(); }
 
-  void addLocation(FeatureSourceRange Fsr) { Locations.push_back(Fsr); }
+  void addLocation(FeatureSourceRange Fsr) {
+    Locations.push_back(std::move(Fsr));
+  }
+
   std::vector<FeatureSourceRange>::iterator
   removeLocation(const FeatureSourceRange &Fsr) {
     return Locations.erase(std::find(Locations.begin(), Locations.end(), Fsr));
--- a/include/vara/Feature/Feature.h
+++ b/include/vara/Feature/Feature.h
@@ -274,4 +274,16 @@ public:
 
 } // namespace vara::feature
 
+inline std::ostream &operator<<(std::ostream &Out,
+                                const vara::feature::Feature *Feature) {
+  Out << Feature->toString();
+  return Out;
+}
+
+inline llvm::raw_ostream &operator<<(llvm::raw_ostream &Out,
+                                     const vara::feature::Feature *Feature) {
+  Out << Feature->toString();
+  return Out;
+}
+
 #endif // VARA_FEATURE_FEATURE_H
--- a/include/vara/Feature/Feature.h
+++ b/include/vara/Feature/Feature.h
@@ -276,14 +276,14 @@ public:
 } // namespace vara::feature
 
 inline std::ostream &operator<<(std::ostream &Out,
-                                const vara::feature::Feature *Feature) {
-  Out << Feature->toString();
+                                const vara::feature::Feature &Feature) {
+  Out << Feature.toString();
   return Out;
 }
 
 inline llvm::raw_ostream &operator<<(llvm::raw_ostream &Out,
-                                     const vara::feature::Feature *Feature) {
-  Out << Feature->toString();
+                                     const vara::feature::Feature &Feature) {
+  Out << Feature.toString();
   return Out;
 }
 
--- a/include/vara/Feature/FeatureModel.h
+++ b/include/vara/Feature/FeatureModel.h
@@ -58,7 +58,7 @@ public:
 
   [[nodiscard]] llvm::StringRef getCommit() const { return Commit; }
 
-  [[nodiscard]] Feature *getRoot() const { return Root; }
+  [[nodiscard]] RootFeature *getRoot() const { return Root; }
 
   //===--------------------------------------------------------------------===//
   // Ordered feature iterator
@@ -131,7 +131,7 @@ private:
   /// Delete a \a Feature.
   void removeFeature(Feature &Feature);
 
-  RootFeature *setRoot(std::unique_ptr<RootFeature> NewRoot);
+  RootFeature *setRoot(RootFeature &NewRoot);
 
   OrderedFeatureTy OrderedFeatures;
 };
--- a/include/vara/Feature/FeatureModel.h
+++ b/include/vara/Feature/FeatureModel.h
@@ -34,8 +34,7 @@ public:
   using OrderedFeatureTy = OrderedFeatureVector;
   using ConstraintTy = Constraint;
   using ConstraintContainerTy = std::vector<std::unique_ptr<ConstraintTy>>;
-  using RelationshipTy = Relationship;
-  using RelationshipContainerTy = std::vector<std::unique_ptr<RelationshipTy>>;
+  using RelationshipContainerTy = std::vector<std::unique_ptr<Relationship>>;
 
   FeatureModel() = default;
   FeatureModel(std::string Name, fs::path RootPath, std::string Commit,
@@ -126,6 +125,11 @@ private:
   ///
   /// \returns ptr to inserted \a Feature
   Feature *addFeature(std::unique_ptr<Feature> Feature);
+
+  Relationship *addRelationship(std::unique_ptr<Relationship> Relationship) {
+    Relationships.push_back(std::move(Relationship));
+    return Relationships.back().get();
+  }
 
   /// Delete a \a Feature.
   void removeFeature(Feature &Feature);
--- a/include/vara/Feature/FeatureModel.h
+++ b/include/vara/Feature/FeatureModel.h
@@ -360,8 +360,7 @@ struct ExactlyOneRootNode {
          }))) {
       return true;
     }
-    llvm::errs() << "Failed to validate 'CheckFeatureParentChildRelationShip'."
-                 << '\n';
+    llvm::errs() << "Failed to validate 'ExactlyOneRootNode'." << '\n';
     return false;
   }
 };
--- a/include/vara/Feature/FeatureModel.h
+++ b/include/vara/Feature/FeatureModel.h
@@ -25,6 +25,7 @@ class FeatureModelModification;
 
 /// \brief Tree like representation of features and dependencies.
 class FeatureModel {
+  friend class FeatureModelBuilder;
   // Only Modifications are allowed to edit a FeatureModel after creation.
   friend class detail::FeatureModelModification;
 
@@ -34,14 +35,14 @@ public:
   using ConstraintContainerTy = std::vector<std::unique_ptr<ConstraintTy>>;
   using RelationshipContainerTy = std::vector<std::unique_ptr<Relationship>>;
 
-  FeatureModel() = default;
-  FeatureModel(std::string Name, fs::path RootPath, std::string Commit,
-               FeatureMapTy Features, ConstraintContainerTy Constraints,
-               RelationshipContainerTy Relationships, RootFeature *Root)
-      : Name(std::move(Name)), Path(std::move(RootPath)),
-        Commit(std::move(Commit)), Features(std::move(Features)),
-        Constraints(std::move(Constraints)),
-        Relationships(std::move(Relationships)), Root(Root) {}
+  FeatureModel(
+      std::string Name = "FeatureModel",
+      std::unique_ptr<RootFeature> Root = std::make_unique<RootFeature>("root"),
+      fs::path Path = "", std::string Commit = "")
+      : Name(std::move(Name)), Root(Root.get()), Path(std::move(Path)),
+        Commit(std::move(Commit)) {
+    addFeature(std::move(Root));
+  }
 
   [[nodiscard]] unsigned int size() { return Features.size(); }
 
@@ -233,12 +234,12 @@ public:
 
 protected:
   std::string Name;
+  RootFeature *Root;
   fs::path Path;
   std::string Commit;
   FeatureMapTy Features;
   ConstraintContainerTy Constraints;
   RelationshipContainerTy Relationships;
-  RootFeature *Root{nullptr};
 
 private:
   /// Insert a \a Feature into existing model.
@@ -471,13 +472,12 @@ struct CheckFeatureParentChildRelationShip {
 
 struct ExactlyOneRootNode {
   static bool check(FeatureModel &FM) {
-    if ((!FM.getRoot() && FM.size() == 0) ||
-        (llvm::isa_and_nonnull<RootFeature>(FM.getRoot()) &&
-         1 == std::accumulate(FM.unordered_features().begin(),
-                              FM.unordered_features().end(), 0,
-                              [](int Sum, Feature *F) {
-                                return Sum + llvm::isa<RootFeature>(F);
-                              }))) {
+    if (llvm::isa_and_nonnull<RootFeature>(FM.getRoot()) &&
+        1 == std::accumulate(FM.unordered_features().begin(),
+                             FM.unordered_features().end(), 0,
+                             [](int Sum, Feature *F) {
+                               return Sum + llvm::isa<RootFeature>(F);
+                             })) {
       return true;
     }
     llvm::errs() << "Failed to validate 'ExactlyOneRootNode'." << '\n';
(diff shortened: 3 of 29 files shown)


include/vara/Feature/Constraint.h
```diff
@@ -595,7 +599,7 @@ class PrimaryFeatureConstraint : public PrimaryConstraint {
   void accept(ConstraintVisitor &V) override;
 
 private:
-  friend FeatureModelBuilder;
+  friend detail::FeatureModelModification;
```
vulder: please move to top of the class

include/vara/Feature/Feature.h
```diff
@@ -274,4 +275,16 @@ class RootFeature : public Feature {
 
 } // namespace vara::feature
 
+inline std::ostream &operator<<(std::ostream &Out,
+                                const vara::feature::Feature *Feature) {
```
```suggestion
                                const vara::feature::Feature &Feature) {
```
Should take the feature via `&` because we cannot handle a null state in the function and don't need to reassign the value.

include/vara/Feature/Feature.h
```diff
@@ -274,4 +275,16 @@ class RootFeature : public Feature {
...
+  Out << Feature->toString();
+  return Out;
+}
+
+inline llvm::raw_ostream &operator<<(llvm::raw_ostream &Out,
+                                     const vara::feature::Feature *Feature) {
```
```suggestion
                                     const vara::feature::Feature &Feature) {
```

include/vara/Feature/FeatureModel.h
```diff
@@ -34,12 +34,12 @@ class FeatureModel {
   using OrderedFeatureTy = OrderedFeatureVector;
   using ConstraintTy = Constraint;
   using ConstraintContainerTy = std::vector<std::unique_ptr<ConstraintTy>>;
-  using RelationshipTy = Relationship;
-  using RelationshipContainerTy = std::vector<std::unique_ptr<RelationshipTy>>;
+  using RelationshipContainerTy = std::vector<std::unique_ptr<Relationship>>;
 
+  FeatureModel() = default;
```
vulder: Why do we need a public default ctor for FM?
The default constructed FM would not make sense because key values are empty.

include/vara/Feature/FeatureModel.h
```diff
@@ -58,9 +58,10 @@ class FeatureModel {
 
   [[nodiscard]] llvm::StringRef getCommit() const { return Commit; }
 
-  void setCommit(std::string Str) { Commit = std::move(Str); }
+  [[nodiscard]] RootFeature *getRoot() const { return Root; }
 
-  [[nodiscard]] Feature *getRoot() const { return Root; }
+  // TODO make this private and use a transaction
```
vulder: TODO

include/vara/Feature/FeatureModel.h
```diff
@@ -130,149 +129,31 @@ class FeatureModel {
...
+  void setName(std::string NewName) { Name = std::move(NewName); }
 
-  bool buildConstraints();
+  void setPath(fs::path NewPath) { Path = std::move(NewPath); }
 
-  /// This method is solely relevant for parsing XML, as alternatives are
-  /// represented als mutual excluded but non-optional features (which requires
-  /// additional processing).
-  void detectXMLAlternatives();
+  void sort() { OrderedFeatures.sort(); }
```
vulder: Is this ever needed? Shouldn't the OrderedFeatureVector guarantee to be always ordered?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -19,10 +20,14 @@ class FeatureModelCopyTransactionBase;
...
 
 using ConsistencyCheck =
-    FeatureModelConsistencyChecker<EveryFeatureRequiresParent,
+    FeatureModelConsistencyChecker<ExactlyOneRootNode,
+                                   EveryFeatureRequiresParent,
                                    CheckFeatureParentChildRelationShip>;
 } // namespace detail
 
+using FeatureVariant = std::variant<std::string, Feature *>;
+using FeatureTreeNodeVariant = std::variant<std::string, FeatureTreeNode *>;
```
vulder: What is the idea behind this 2 global var defs?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -58,7 +63,12 @@ class FeatureModelTransaction
              "Transaction in CopyMode should be commited before destruction.");
     } else {
       if (this->isUncommited()) { // In modification mode we should ensure that
-                                  // changes are commited before destruction
+                                  // changes are committed before destruction
+        llvm::errs()
+            << "warning: Uncommitted modifications before destruction.\n";
+        // TODO(s9latimm): Committing now may break with prior failed commits.
```
vulder: Set in on the agenda for friday

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -163,6 +255,14 @@ class FeatureModelModification {
     FM.removeFeature(F);
   }
 
+  template <class PtrTy>
+  static PtrTy *resolveVariant(FeatureModel &FM,
+                               std::variant<std::string, PtrTy *> V) {
+    return std::holds_alternative<std::string>(V)
+               ? FM.getFeature(std::get<std::string>(V))
+               : std::get<PtrTy *>(V);
+  }
```
vulder: The name is a bit confusing and unspecific.
I also think, this method is not necessary. If anyone needs to handle a variant he should use `visit` to handle the different types.

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -205,15 +308,235 @@ class AddFeatureToModel : public FeatureModelModification {
...
+  void exec(FeatureModel &FM) override { (*this)(FM); }
+
+  Relationship *operator()(FeatureModel &FM) {
+    auto *InsertedRelationship =
+        addRelationship(FM, std::make_unique<Relationship>(Kind));
+    if (!InsertedRelationship) {
+      return nullptr;
+    }
+
+    auto *P = resolveVariant(FM, Parent);
```
vulder: This should use visit

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -205,15 +308,235 @@ class AddFeatureToModel : public FeatureModelModification {
...
+//                              SetName
+//===----------------------------------------------------------------------===//
+
+class SetName : public FeatureModelModification {
+  friend class FeatureModelModification;
+
+public:
+  void exec(FeatureModel &FM) override { (*this)(FM); }
+
+  void operator()(FeatureModel &FM) { setName(FM, Name); }
```
vulder: You could move in the name

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -205,15 +308,235 @@ class AddFeatureToModel : public FeatureModelModification {
...
+//                              SetCommit
+//===----------------------------------------------------------------------===//
+
+class SetCommit : public FeatureModelModification {
+  friend class FeatureModelModification;
+
+public:
+  void exec(FeatureModel &FM) override { (*this)(FM); }
+
+  void operator()(FeatureModel &FM) { setCommit(FM, Commit); }
```
vulder: same with commit

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -205,15 +308,235 @@ class AddFeatureToModel : public FeatureModelModification {
...
+//                              SetPath
+//===----------------------------------------------------------------------===//
+
+class SetPath : public FeatureModelModification {
+  friend class FeatureModelModification;
+
+public:
+  void exec(FeatureModel &FM) override { (*this)(FM); }
+
+  void operator()(FeatureModel &FM) { setPath(FM, Path); }
```
vulder: same with path

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -205,15 +308,235 @@ class AddFeatureToModel : public FeatureModelModification {
...
+        for (auto *C : FM.getRoot()->children()) {
+          setParent(*C, *NewRoot);
+          removeEdge(*FM.getRoot(), *C);
+          addEdge(*NewRoot, *C);
+        }
+        removeFeature(FM, *FM.getRoot());
+      }
+      setRoot(FM, *NewRoot);
+    }
+    sort(FM);
```
vulder: Why is an explicit sort call needed here?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -205,15 +308,235 @@ class AddFeatureToModel : public FeatureModelModification {
...
+
+class AddChild : public FeatureModelModification {
+  friend class FeatureModelModification;
+
+public:
+  void exec(FeatureModel &FM) override { (*this)(FM); }
+
+  void operator()(FeatureModel &FM) {
+    auto *C = resolveVariant(FM, Child);
+    auto *P = resolveVariant(FM, Parent);
```
vulder: Visit

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -205,15 +308,235 @@ class AddFeatureToModel : public FeatureModelModification {
...
+    removeEdge(*C->getParent(), *C);
+    addEdge(*P, *C);
+    setParent(*C, *P);
+    sort(FM);
```
vulder: ?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -232,35 +555,101 @@ class FeatureModelCopyTransactionBase {
...
+                           const FeatureTreeNodeVariant &Child) {
+    FeatureModelModification::make_modification<AddChild>(Parent, Child);
+  }
+
 private:
   std::unique_ptr<FeatureModel> FM;
+
+  [[nodiscard]] Feature *TranslateFeature(Feature &F) {
+    return FM->getFeature(F.getName());
+  }
```
```suggestion
  [[nodiscard]] Feature *TranslateFeature(Feature &F) {
    return FM->getFeature(F.getName());
  }
  
  std::unique_ptr<FeatureModel> FM;
```
Member vars should be last

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -278,13 +667,97 @@ class FeatureModelModifyTransactionBase {
...
+                            AddLocationToFeature>(F, std::move(FSR)));
+  }
+
+  void addConstraintImpl(std::unique_ptr<Constraint> NewConstraint) {
+    Modifications.push_back(FeatureModelModification::make_unique_modification<
+                            AddConstraintToModel>(std::move(NewConstraint)));
+  }
+
+  void setNameImpl(std::string Name) {
+    assert(FM && "");
```
vulder: We should fill the asserts with an useful message

include/vara/Feature/OrderedFeatureVector.h
```diff
@@ -57,6 +59,13 @@ class OrderedFeatureVector {
     insert(Init.begin(), Init.end());
   }
 
+  void sort() {
```
vulder: same as mentioned above

lib/Feature/FeatureModelBuilder.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+namespace vara::feature {
+
+//===----------------------------------------------------------------------===//
+//                        FeatureModelBuilder
+//===----------------------------------------------------------------------===//
+
+/// Decide whether two features are mutual exclusive. Beware that this method
+/// only detects very simple trees with binary excludes.
+bool isSimpleMutex(const Feature *A, const Feature *B) {
```
vulder: I think, we should simplify this implementation a bit by refactoring out some parts. This is quite hard to understand.

lib/Feature/FeatureModelBuilder.cpp
```diff
@@ -0,0 +1,89 @@
...
+        std::all_of(Children.begin(), Children.end(), [Children](auto *F) {
+          for (auto *C : Children) {
+            if (F == C) {
+              continue;
+            }
+            if (!isSimpleMutex(F, C)) {
+              return false;
+            }
+          }
+          return !F->isOptional();
```
vulder: This could probability be simplified by using a second `all_of`

lib/Feature/FeatureModelBuilder.cpp
```diff
@@ -0,0 +1,89 @@
...
+std::unique_ptr<FeatureModel> FeatureModelBuilder::buildFeatureModel() {
+  if (!FM->getRoot()) {
+    auto T = FeatureModelModifyTransaction::openTransaction(*FM);
+    T.setRoot(std::make_unique<RootFeature>("root"));
+    if (!T.commit()) {
+      return nullptr;
+    }
+  }
+  return Features.commit() && Transactions.commit() &&
+                 PostTransactions.commit() && detectXMLAlternatives() &&
```
vulder: What are PostTransactions?

lib/Feature/FeatureModelBuilder.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+std::unique_ptr<FeatureModel> FeatureModelBuilder::buildFeatureModel() {
+  if (!FM->getRoot()) {
+    auto T = FeatureModelModifyTransaction::openTransaction(*FM);
+    T.setRoot(std::make_unique<RootFeature>("root"));
+    if (!T.commit()) {
+      return nullptr;
+    }
+  }
+  return Features.commit() && Transactions.commit() &&
```
vulder: What is the meaning of Transactions here?

vulder: Hmm, I think we could just chain the commits sequentially instead of pressing everything into one big && expression

unittests/Feature/FeatureModelBuilder.cpp
```diff
@@ -126,17 +126,19 @@ TEST(FeatureModelBuilder, addNumericFeatureRef) {
...
+  B.makeFeature<BinaryFeature>("aa", false)->addEdge("a", "aa");
+  B.makeFeature<BinaryFeature>("ab", false)->addEdge("a", "ab");
 
   B.addConstraint(std::make_unique<ExcludesConstraint>(
-      std::make_unique<PrimaryFeatureConstraint>(AA),
-      std::make_unique<PrimaryFeatureConstraint>(AB)));
+      std::make_unique<PrimaryFeatureConstraint>(
+          std::make_unique<BinaryFeature>("aa")),
+      std::make_unique<PrimaryFeatureConstraint>(
+          std::make_unique<BinaryFeature>("ab"))));
```
vulder: Maybe a `using namespace std;` would make sense at the top of this file :wink:

unittests/Feature/FeatureModelBuilder.cpp
```diff
@@ -157,47 +160,102 @@ TEST(FeatureModelBuilder, detectXMLAlternativesBroken) {
...
+  B.addConstraint(std::make_unique<ExcludesConstraint>(
+      std::make_unique<PrimaryFeatureConstraint>(
+          std::make_unique<BinaryFeature>("ab")),
+      std::make_unique<PrimaryFeatureConstraint>(
+          std::make_unique<BinaryFeature>("ac"))));
+  B.addConstraint(std::make_unique<ExcludesConstraint>(
+      std::make_unique<PrimaryFeatureConstraint>(
+          std::make_unique<BinaryFeature>("aa")),
+      std::make_unique<PrimaryFeatureConstraint>(
+          std::make_unique<BinaryFeature>("ac"))));
```
vulder: This pattern looks repetitive, I think we should create a helper/short cut for this.

vulder (changes requested):
What a big rewrite :laughing: 
But I guess the new design works?

include/vara/Feature/FeatureModel.h
s9latimm: This is kind of required, if we want to build a FM only using transactions, so we start with an empty tree.
It may be reasonable to initialize a root feature at this point, which would make that explicit step in the builder obsolete.
Currently completely empty FM (without any nodes and especially without a root) are valid.

include/vara/Feature/FeatureModel.h
s9latimm: The ordering inside the vector is currently based on the hierarchy of a feature at the moment it is inserted.
So this is currently not dynamically updated.
As we can change the structure of the FM through transaction, we need to reevaluate this sometimes (for example based on changed parents).

I think we should provide this constraint with the collection itself -- will look into it.

include/vara/Feature/FeatureModelTransaction.h
s9latimm: I found it suitable to provide transactions with the ability to handle features both based on name or pointer.
In most cases the provided feature may not exist yet and only be created while committing, so we can only refer to it by name -- not sure if supporting pointers has huge benefits.

include/vara/Feature/FeatureModelTransaction.h
s9latimm: This may be obsolete, if we decide that transactions only take names of features and not pointers (see above).

lib/Feature/FeatureModelBuilder.cpp
s9latimm: These are dependent stages, if we want to assure, all feature reference can be resolved.

1. `Features` are all emplacements of feature nodes into the FM, without any structural information yet -- basically everything is child of root.
2. `Transactions` contains all modification to build the tree structure of features, like parent / child relations.
3. `PostTransactions` (we may use a better name, see above) insert relationships an constraints. At this stage we can assume, that all feature nodes and edges are already present, so we can easily change the hierarchy or let constraints refer to existing features (through this could also be done in the prior stage).
4. `Special` currently only handles XML alternative groups, so we could move this locally, but I think natively providing a collection for similar corner cases might be useful.

include/vara/Feature/FeatureModel.h
```diff
@@ -443,30 +324,45 @@ template <typename... Rules> class FeatureModelConsistencyChecker {
...
-                                  return Sum + llvm::isa<RootFeature>(F);
-                                });
+    if ((!FM.getRoot() && FM.size() == 0) ||
+        (llvm::isa_and_nonnull<RootFeature>(FM.getRoot()) &&
+         1 == std::accumulate(FM.begin(), FM.end(), 0, [](int Sum, Feature *F) {
+           return Sum + llvm::isa<RootFeature>(F);
+         }))) {
+      return true;
+    }
+    llvm::errs() << "Failed to validate 'CheckFeatureParentChildRelationShip'."
```
padupr:
```suggestion
    llvm::errs() << "Failed to validate 'ExactlyOneRootNode'."
```

include/vara/Feature/FeatureModel.h
s9latimm: I think making the FMbuilder a friend and this private may be a better approach.

We might want to support a constructor without features, constraints etc though.

unittests/Feature/FeatureModelBuilder.cpp
s9latimm: I added a macro. 
With this the count of `std::` is significantly reduced, so I think we do not need to add `using namespace std` here.

vulder: Please don't use macros :cry:, it should be possible to just build is as a constexpr function or maybe a template func.

include/vara/Feature/FeatureModelTransaction.h
vulder: @s9latimm What do we do here?

lib/Feature/FeatureModel.cpp
```diff
@@ -20,26 +25,25 @@ Feature *FeatureModel::addFeature(std::unique_ptr<Feature> NewFeature) {
...
+
 std::unique_ptr<FeatureModel> FeatureModel::clone() {
   FeatureModelBuilder FMB;
   FMB.setVmName(this->getName().str());
   FMB.setCommit(this->getCommit().str());
   FMB.setPath(this->getPath().string());
-  FMB.setRootName(this->getRoot()->getName().str());
 
   for (const auto &KV : this->Features) {
     // TODO(s9latimm): Add unittests for cloned FeatureSourceRanges
```
vulder: TODO

vulder (approved):

padupr (approved):

include/vara/Feature/FeatureModelTransaction.h
s9latimm: This strongly depends on se-passau/VaRA/issues/744, as we need a way to track if a prior commit failed to prevent breaking the FM further and prohibit unintended behavior -- this may be combined with a rollback (see se-passau/VaRA/issues/723) while discarding any uncommitted changes or even require success when invoked without a prior commit.

I kept this open for now and linked the related issue.

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -63,13 +63,14 @@ class FeatureModelTransaction
...
-                                   // changes are committed before destruction
+      if (this->isUncommitted()) {
+        // In modification mode we should ensure that changes are committed
+        //  before destruction
         llvm::errs()
             << "warning: Uncommitted modifications before destruction.\n";
-        // TODO(s9latimm): Committing now may break with prior failed commits.
-        //  We need a better way of tracking if a previous commit failed,
-        //  instead of checking locally if FM is nullptr.
+        // TODO(se-passau/VaRA/issues/744): Committing now may break with prior
```
s9latimm:
```suggestion
        // TODO(se-passau/VaRA#744): Committing now may break with prior
```

PR #44, opened 2021-03-09
Merging Feature Models

padupr:


include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -112,6 +121,17 @@ using FeatureModelModifyTransaction =
 void addFeature(FeatureModel *FM, std::unique_ptr<Feature> NewFeature,
                 Feature *Parent = nullptr);
 
+/// Merges a FeatureModel into another
+///
+/// Merging fails if both FeatureModels contain a Feature with equal name,
+/// but different properties.
+///
+/// \param FM1
+/// \param FM2
```
```suggestion
/// \param FM1
/// \param FM2
///
```

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -133,16 +153,23 @@ class FeatureModelModification {
 
 protected:
   /// \brief Set the parent of a \a Feature.
-  static void setParent(Feature &F, Feature *Parent) { F.setParent(Parent); }
+  static void setParent(FeatureTreeNode &F, FeatureTreeNode *Parent) {
+    // TODO Discuss: should this be set to the child relationship if it exists?
```
vulder: What do you mean with this, the `FeatureTreeNode`?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -205,6 +238,42 @@ class AddFeatureToModel : public FeatureModelModification {
...
+    }*/
+
+    setParent(*RS, GroupRoot);
+    addChild(*GroupRoot, *RS);
+    for (auto *Child : GroupRoot->getChildren<Feature>()) {
+      removeChild(*GroupRoot, *Child);
+      addChild(*RS, *Child);
+      setParent(*Child, RS);
+    }
+    return GroupRoot;
```
vulder: What is the idea behind returning `GroupRoot`, should the caller not already have this, because he created the Mod. with the `GroupRoot`?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -241,8 +310,22 @@ class FeatureModelCopyTransactionBase {
...
+    }
+
+    return FeatureModelModification::make_modification<
+        AddRelationshipToFeature>(Kind, TranslateFeature(GroupRoot))(*FM);
+  }
+
 private:
   std::unique_ptr<FeatureModel> FM;
+
+  Feature *TranslateFeature(const Feature *F) const {
```
```suggestion
  Feature *MapToCopiedFM(const Feature *F) const {
```
It think we should call the method something with `map`, about the Copied part I'm not so sure^^

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -4,11 +4,111 @@
...
+  assert(F1.getName() == F2.getName());
+  if (F1.getKind() != F2.getKind()) {
+    return false;
+  }
+  if (F1.isOptional() != F2.isOptional()) {
+    return false;
+  }
+  if (F1.getKind() == Feature::FeatureKind::FK_ROOT) {
+    return true;
+  }
```
vulder, line 96: Hmm, maybe I don't understand this soft compare but why are they equal if F1 is a `FK_ROOT`?

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -4,11 +4,111 @@
...
+  std::unique_ptr<Feature> FeatureSoftCopy;
+  NumericFeature::ValuesVariantType Values;
+  switch (F->getKind()) {
+  case Feature::FeatureKind::FK_BINARY:
+    return std::make_unique<BinaryFeature>(
+        F->getName().str(), F->isOptional(),
+        std::vector<FeatureSourceRange>(F->getLocationsBegin(),
+                                        F->getLocationsEnd()));
+  case Feature::FeatureKind::FK_NUMERIC:
+    Values = dynamic_cast<NumericFeature *>(F)->getValues();
```
vulder: For the llvm based types you need `dyn_cast`

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -4,11 +4,111 @@
...
+
+[[nodiscard]] std::unique_ptr<Feature> FeatureSoftCopy(Feature *F) {
+  std::unique_ptr<Feature> FeatureSoftCopy;
+  NumericFeature::ValuesVariantType Values;
+  switch (F->getKind()) {
+  case Feature::FeatureKind::FK_BINARY:
+    return std::make_unique<BinaryFeature>(
+        F->getName().str(), F->isOptional(),
+        std::vector<FeatureSourceRange>(F->getLocationsBegin(),
+                                        F->getLocationsEnd()));
```
vulder: What do you mean by soft copy here? For me this looks like a deep copy of a feature.

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -4,11 +4,111 @@
...
+  auto Trans = FeatureModelCopyTransaction::openTransaction(FM1);
+  if (!mergeSubtree(Trans, FM1, *FM2.getRoot())) {
+    return nullptr;
+  } else {
+    return Trans.commit();
+  }
+}
+
+[[nodiscard]] bool mergeSubtree(FeatureModelCopyTransaction &Trans,
+                                FeatureModel &FM, Feature &F) {
```
vulder: FM should not be modifiable here I guess? So, we could pass it as a `const &`

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -4,11 +4,111 @@
...
+        }
+      }
+    } else {
+      Trans.abort();
+      return false;
+    }
+  } else {
+    std::unique_ptr<Feature> SoftCopy = FeatureSoftCopy(&F);
+    if (!SoftCopy) {
+      Trans.abort();
```
vulder: Hmm, maybe it would be easier to handle aborting the transaction on the top level.
Just return false here and in the case above before you return a nullptr abort the transaction

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -4,11 +4,111 @@
...
+  }
+}
+
+[[nodiscard]] bool mergeSubtree(FeatureModelCopyTransaction &Trans,
+                                FeatureModel &FM, Feature &F) {
+  // Is there a similar Feature in the other FM
+  if (Feature *CMP = FM.getFeature(F.getName())) {
+    if (FeatureSoftCompare(*CMP, F)) {
+      // similar feature, maybe merge locations
+      for (FeatureSourceRange &FSR : F.getLocations()) {
```
vulder: `const`?

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+  }
+
+  std::unique_ptr<FeatureModel> FM;
+};
+
+TEST_F(FeatureModelMergeTransactionTest, Simple) {
+  FeatureModelBuilder B;
+  B.makeFeature<BinaryFeature>("b", true);
+  auto FM2 = B.buildFeatureModel();
+  assert(FM);
```
```suggestion
  assert(FM2);
```

vulder: Also, you could just use the ASSERT from gtest here.

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+  EXPECT_EQ(Feature::FeatureKind::FK_ROOT, F2->getParentFeature()->getKind());
+}
+
+TEST_F(FeatureModelMergeTransactionTest, Idempotence) {
+  size_t FMSizeBefore = FM->size();
+
+  auto FMMerged = mergeFeatureModels(*FM, *FM);
+  assert(FMMerged);
+
+  EXPECT_EQ(FMSizeBefore, FMMerged->size());
```
vulder, line 267: add a check here if it's sill 'a' with the correct type

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+  B.makeFeature<BinaryFeature>("a", true,
+                               std::vector<FeatureSourceRange>{FSR1, FSR3});
+  auto FM2 = B.buildFeatureModel();
+  assert(FM2);
+
+  auto FMMerged = mergeFeatureModels(*FM, *FM2);
+  assert(FMMerged);
+
+  Feature *F = FM->getFeature("a");
+  EXPECT_EQ(3, std::distance(F->getLocationsBegin(), F->getLocationsEnd()));
```
vulder, line 291: Hmm, I would expect that locs is `2` here. FSR2 and FSR3 are the same, so they should be "merged" because they carry the same information.
In the case that the location differ, I would agree we should have all 3 afterwards.

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+  size_t FMSizeBefore = FM->size();
+
+  FeatureModelBuilder B;
+  B.makeFeature<BinaryFeature>("b", true);
+  B.addEdge("b", "ba");
+  B.addEdge("b", "bb");
+  B.makeFeature<BinaryFeature>("ba", false);
+  B.makeFeature<BinaryFeature>("bb", false);
+  auto FM2 = B.buildFeatureModel();
+  assert(FM);
```
```suggestion
  assert(FM2);
```

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+
+  auto FMMerged = mergeFeatureModels(*FM, *FM2);
+  assert(FMMerged);
+
+  EXPECT_EQ(FMSizeBefore + 3, FMMerged->size());
+  Feature *FA = FMMerged->getFeature("a");
+  EXPECT_TRUE(FA);
```
```suggestion
  EXPECT_TRUE(FMMerged->getFeature("a"));
```

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+
+  EXPECT_EQ("b", FBA->getParentFeature()->getName());
+  EXPECT_EQ("b", FBB->getParentFeature()->getName());
+}
+
+TEST_F(FeatureModelMergeTransactionTest, RejectDifferenceOptional) {
+  FeatureModelBuilder B;
+  B.makeFeature<BinaryFeature>("a", false);
+  auto FM2 = B.buildFeatureModel();
+  assert(FM);
```
```suggestion
  assert(FM2);
```

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+
+  // Expect fail, property optional is different
```
vulder, line 332: Why can we not merge optional differences?

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+  EXPECT_FALSE(FMMerged);
+}
+
+TEST_F(FeatureModelMergeTransactionTest, RejectDifferenceParent) {
+  FeatureModelBuilder B;
+  B.makeFeature<BinaryFeature>("b", true);
+  B.addEdge("b", "a");
+  B.makeFeature<BinaryFeature>("a", true);
+  auto FM2 = B.buildFeatureModel();
+  assert(FM);
```
```suggestion
  assert(FM2);
```

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+
+  // Expect fail, feature a has different parents; root vs. b
```
vulder, line 345: Could we not handle this by looking up the other parent, or creating the correct parent?

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+  auto FMMerged = mergeFeatureModels(*FM, *FM2);
+  EXPECT_FALSE(FMMerged);
+}
+
+TEST_F(FeatureModelMergeTransactionTest, RejectDifferenceKind) {
+  FeatureModelBuilder B;
+  B.makeFeature<NumericFeature>("a", std::vector<int>{1, 2, 3}, true);
+  auto FM2 = B.buildFeatureModel();
+  assert(FM);
```
```suggestion
  assert(FM2);
```

vulder (changes requested):

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -4,11 +4,111 @@
...
 }
 
+[[nodiscard]] std::unique_ptr<FeatureModel>
+mergeFeatureModels(FeatureModel &FM1, FeatureModel &FM2) {
+  auto Trans = FeatureModelCopyTransaction::openTransaction(FM1);
+  if (!mergeSubtree(Trans, FM1, *FM2.getRoot())) {
+    return nullptr;
+  } else {
+    return Trans.commit();
+  }
```
```suggestion
  }
  return Trans.commit();
```

lib/Feature/FeatureModelTransaction.cpp
padupr: I checked before that both have the same kind. Root features do not have any other meaningful properties. Maybe we should also ignore the name for root features

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -224,4 +224,238 @@ TEST_F(FeatureModelTransactionModifyTest, addFeatureToModelThenAboard) {
...
+  EXPECT_EQ(FMSizeBefore, FMMerged->size());
+}
+
+TEST_F(FeatureModelMergeTransactionTest, DifferentLocations) {
+  FeatureSourceRange::FeatureSourceLocation FSL1(2, 4);
+  FeatureSourceRange::FeatureSourceLocation FSL2(2, 30);
+  FeatureSourceRange FSR1("path", FSL1, FSL2);
+  // TODO there must be a way to pass a copy directly
+  FeatureSourceRange FSRC1(FSR1);
+  FM->getFeature("a")->addLocation(FSRC1);
```
padupr: @vulder is there a way to pass the FeatureSourceRange here directly without saving it in a local variable before?

unittests/Feature/FeatureModelTransaction.cpp
padupr: Forgot to change FSR3, they should not be equal, thus 3 locations.

unittests/Feature/FeatureModelTransaction.cpp
padupr: As discussed I will implement different merging strategies in the future.

unittests/Feature/FeatureModelTransaction.cpp
padupr: I think this would be as simple as to only add the missing features. If the parents are different, the feature is already in the final model. The parent cannot be missing since it is be merged before. Still need to think about possible edge cases.

vulder: kk

unittests/Feature/FeatureModelTransaction.cpp
vulder: ok

lib/Feature/FeatureModelTransaction.cpp
vulder: Ah, I see. You check for root here because all root related properties have already been checked. :+1:

unittests/Feature/FeatureModelTransaction.cpp
vulder: Oh, boy. Not with the current impl. Just change the function `addLocation(FeatureSourceRange FSR)` without the `&`. The std::move() afterwards also makes not much sense in the current impl with & :laughing:

Then you can just use it like this:
```cpp
    FM->getFeature("a")->addLocation(FeatureSourceRange{"path", {2, 4}, {2, 30}});
```

include/vara/Feature/Feature.h
```diff
@@ -92,9 +92,7 @@ class Feature : public FeatureTreeNode {
   // Locations
   [[nodiscard]] bool hasLocations() const { return !Locations.empty(); }
 
-  void addLocation(FeatureSourceRange &Fsr) {
-    Locations.push_back(std::move(Fsr));
-  }
+  void addLocation(FeatureSourceRange Fsr) { Locations.push_back(Fsr); }
```
```suggestion
  void addLocation(FeatureSourceRange Fsr) { Locations.push_back(std::move(Fsr)); }
```
For the value type FSR we need a move. Only before with the ref it was not needed

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -95,12 +95,46 @@ class FeatureModelTransaction
...
   }
 
-  decltype(auto) addRelationship(Relationship::RelationshipKind Kind,
-                                 Feature *GroupRoot) {
+  decltype(auto) addRelationship(
+      Relationship::RelationshipKind Kind, const FeatureVariant &Parent,
+      std::variant<std::set<std::string>, std::set<Feature *>> Children) {
     if constexpr (IsCopyMode) {
-      return this->addRelationshipImpl(Kind, GroupRoot);
+      return this->addRelationshipImpl(Kind, Parent, Children);
```
```suggestion
      return this->addRelationshipImpl(Kind, Parent, std::move(Children));
```

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -95,12 +95,46 @@ class FeatureModelTransaction
...
     } else {
-      this->addRelationshipImpl(Kind, GroupRoot);
+      this->addRelationshipImpl(Kind, Parent, Children);
```
```suggestion
      this->addRelationshipImpl(Kind, Parent, std::move(Children));
```

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -274,15 +274,12 @@ TEST_F(FeatureModelMergeTransactionTest, DifferentLocations) {
   FeatureSourceRange::FeatureSourceLocation FSL1(2, 4);
   FeatureSourceRange::FeatureSourceLocation FSL2(2, 30);
   FeatureSourceRange FSR1("path", FSL1, FSL2);
-  // TODO there must be a way to pass a copy directly
-  FeatureSourceRange FSRC1(FSR1);
-  FM->getFeature("a")->addLocation(FSRC1);
+  FM->getFeature("a")->addLocation(FSR1);
```
vulder, line 275: I think we wanted to rewrite this to directly create the FSR

vulder (changes requested):

unittests/Feature/FeatureModelTransaction.cpp
vulder: I fixed this.

vulder (approved):

PR #46, opened 2021-04-19
Updated Pythonbindings

padupr:
Updated python bindings to allow basic manipulations of feature models.


bindings/python/vara-feature/pybind_FeatureModel.cpp
```diff
@@ -35,6 +35,27 @@ void init_feature_model_module(py::module &M) {
...
+      .def(
+          "add_feature",
+          [](vf::FeatureModel &FM, vf::Feature &NewFeature,
+             vf::Feature &Parent) {
+            // TODO. Pybind does not allow to pass a unique pointer into
+            // functions, since python has no concept of releasing ownership
+            // (https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html).
+            // While there is some work in progress to allow this, for now we
+            // can either create a copy of the Feature or make a custom smart
+            // pointer. vf::addFeature(FM, std::move(NewFeature), &Parent);
```
padupr: Here we need to transfer ownership from python to c++.

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -16,6 +16,27 @@ void addFeature(FeatureModel &FM, std::unique_ptr<Feature> NewFeature,
...
+  auto Trans = FeatureModelModifyTransaction::openTransaction(FM);
+  detail::FeatureVariantTy F = detail::FeatureVariantTy(OldFeature);
+  Trans.removeFeature(F, Recursive);
+  Trans.commit();
+}
+
+void addRelationship(FeatureModel &FM, Feature *GroupRoot,
+                     Relationship::RelationshipKind Kind) {
+  auto Trans = FeatureModelModifyTransaction::openTransaction(FM);
+  detail::FeatureVariantTy GR = detail::FeatureVariantTy(GroupRoot);
```
MiriamMarx: Can you already provide the variant as a param to the function?

padupr: We could do that. However, if we do, we should use them consistently in all helper functions.
Originally, the variants were introduced to reference a feature for which you could not provide a pointer. Here, the caller can always get the pointer with FM.getFeature(name).

vulder: That's a good point, I guess if the user has a FM it is better to always require a pointer. Supporting a lookup with just a given name would mean we have to deal with the error handling if it is not present. If we only take a feature we can make sure the user lookup the feature and handled what ever problems he had if he did not find one for his name.

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -760,7 +760,7 @@ class FeatureModelCopyTransactionBase {
                                            Parent)))(*FM);
   }
 
-  void removeRelationshipImpl(FeatureVariantTy &F) {
+  void removeRelationshipImpl(const FeatureVariantTy &F) {
     assert(FM && "");
```
```suggestion
    assert(FM && "FeatureModel was null an can not be used.");
```

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -5,9 +5,9 @@
 namespace vara::feature {
 
 bool mergeSubtree(FeatureModelCopyTransaction &Trans, FeatureModel const &FM,
-                  Feature &F);
+                  Feature &F, bool Strict);
 std::unique_ptr<Feature> FeatureCopy(Feature &F);
```
vulder, line 10:
```suggestion

std::unique_ptr<Feature> FeatureCopy(Feature &F);

```

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -34,27 +34,43 @@ void removeFeature(FeatureModel &FM,
...
+void addRelationship(FeatureModel &FM,
+                     const detail::FeatureVariantTy &GroupRoot,
+                     Relationship::RelationshipKind Kind) {
+  auto Trans = FeatureModelModifyTransaction::openTransaction(FM);
+  Trans.addRelationship(Kind, GroupRoot);
+  Trans.commit();
+}
+
+void removeRelationship(FeatureModel &FM,
+                        const detail::FeatureVariantTy &GroupRoot) {
```
vulder, line 48: This would need to change, depending on the variant vs pointer discussion.

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -104,22 +120,31 @@ mergeFeatureModels(FeatureModel &FM1, FeatureModel &FM2) {
...
-    return false;
+  if (F1.getKind() != Feature::FeatureKind::FK_ROOT &&
+      F2.getKind() != Feature::FeatureKind::FK_ROOT) {
+    if (*(F1.getParentFeature()) != *(F2.getParentFeature())) {
+      return false;
+    }
+  }
+  if (!Strict) {
+    // strict merging still requires equal parent names, otherwise we might
+    // introduce implicit constraint due to model structure, e.g. merging group
```
```suggestion
    // introduce implicit constraint due to model structure, e.g., merging group
```

bindings/python/tests/feature_model/test_feature_model.py
```diff
@@ -90,3 +91,96 @@ def test_iter_children(self):
         self.assertTrue(test_feature_root.is_child(test_feature_a))
         self.assertTrue(test_feature_root.is_child(test_feature_b))
         self.assertTrue(test_feature_root.is_child(test_feature_c))
+
+    def test_merge_models_idempotence(self):
+        """ Checks if a feature model merged with itself is unchanged. """
+        test_merged = self.fm.merge_with(self.fm)
+
```
vulder, line 100: maybe add also a before/after size check, so no new "feautres" where added

bindings/python/tests/feature_model/test_feature_model.py
```diff
@@ -90,3 +91,96 @@ def test_iter_children(self):
...
+        self.assertTrue(test_merged.get_feature("N"))
+
+    def test_merge_models(self):
+        """ Checks if merging a feature model with another works correctly. """
+        with open(TEST_INPUTS_DIR / "simple_example_feature_model_merge.xml",
+                  'r') as fm_file:
+            parser = fm_parsers.FeatureModelXmlParser(fm_file.read())
+            test_merge_fm = parser.build_feature_model()
+        test_merged = self.fm.merge_with(test_merge_fm)
+
```
vulder, line 125: Can you make it clear what should change and what should not change. A small comment a separator or maybe adapt the function comment so we know what should happen here.

bindings/python/tests/feature_model/test_feature_model.py
```diff
@@ -90,3 +91,96 @@ def test_iter_children(self):
...
+        self.assertTrue(test_merged.get_feature("C"))
+        self.assertTrue(test_merged.get_feature("N"))
+        self.assertTrue(test_merged.get_feature("X"))
+        self.assertTrue(test_merged.get_feature("Y"))
+        self.assertTrue(test_merged.get_feature("Z"))
+
+
+class TestFeatureModelModifications(unittest.TestCase):
+    """
+    Check if ur generated FeatureModel bindings work correctly.
```
vulder:
```suggestion
    Check if our generated FeatureModel bindings work correctly.
```

bindings/python/tests/feature_model/test_feature_model.py
```diff
@@ -90,3 +91,96 @@ def test_iter_children(self):
...
+    """
+
+    def setUp(cls):
+        """Parse and load a FeatureModel for testing."""
+        # TODO read model once and copy it for each test
```
vulder, line 157: TODO

vulder (changes requested):
Found just a few minor issues.
By the way, did you update the pybind version to a newer one? I see that the submodule changed.

Otherwise, this looks fine.

lib/Feature/FeatureModelTransaction.cpp
padupr: Decided in meeting to use references instead of pointers of FeatureVariant.

include/vara/Feature/FeatureModelTransaction.h
padupr: Also fixed other assertions with empty strings.

bindings/python/tests/feature_model/test_feature_model.py
padupr: Also added size checks for other transaction tests.

bindings/python/tests/feature_model/test_feature_model.py
```diff
@@ -94,7 +94,9 @@ def test_iter_children(self):
 
     def test_merge_models_idempotence(self):
         """ Checks if a feature model merged with itself is unchanged. """
+        test_size = self.fm.size()
```
vulder, line 97:
```suggestion
        test_size = self.fm.size()

```

bindings/python/tests/feature_model/test_feature_model.py
```diff
@@ -94,7 +94,9 @@ def test_iter_children(self):
...
         test_merged = self.fm.merge_with(self.fm)
+        self.assertEqual(test_size, test_merged.size())
```
vulder, line 101:
```suggestion

        self.assertEqual(test_size, test_merged.size())
```

vulder (approved):

PR #49, opened 2021-05-09
F 748 verify feature model function constness

MiriamMarx:


include/vara/Feature/FeatureModel.h
```diff
@@ -168,9 +179,9 @@ class FeatureModel {
     FeatureMapIterator &operator=(FeatureMapIterator &&) = delete;
     ~FeatureMapIterator() = default;
 
-    reference operator*() { return MapIter->getValue().get(); }
+    reference operator*() const { return MapIter->getValue().get(); }
 
-    pointer operator->() { return **this; }
+    pointer operator->() const { return **this; }
```
vulder: The const version for this method should return const_reference and const_pointers.
In addition, we should add non const functions, so it depends on the constness of the FMIterator if accessed values are const or not.

vulder (changes requested):
Otherwise, this PR looks good, thanks :smile:

vulder (approved):

PR #50, opened 2021-05-13
Implements addFeatures and removeFeature for FeatureModelTransaction.

MiriamMarx:
related se-sic/VaRA#749

### DIFF
--- a/include/vara/Feature/FeatureModelTransaction.h
+++ b/include/vara/Feature/FeatureModelTransaction.h
@@ -61,7 +61,7 @@ public:
   ~FeatureModelTransaction() {
     if constexpr (IsCopyMode) {
       assert(!this->isUncommitted() &&
-             "Transaction in CopyMode should be commited before destruction.");
+             "Transaction in CopyMode should be committed before destruction.");
     } else {
       if (this->isUncommitted()) {
         // In modification mode we should ensure that changes are committed
@@ -979,6 +979,25 @@ private:
 /// \param Parent of the new feature
 void addFeature(FeatureModel &FM, std::unique_ptr<Feature> NewFeature,
                 Feature *Parent = nullptr);
+
+/// Adds multiple Features to the FeatureModel
+///
+/// The vector of Features contains tuples of a parent and a new Feature.
+/// If a Parent is passed it needs to be already in the FeatureModel,
+/// otherwise, root is assumed as the parent Feature.
+///
+/// \param FM
+/// \param NewFeatures
+void addFeatures(
+    FeatureModel &FM,
+    std::vector<std::tuple<std::unique_ptr<Feature>, Feature *>> NewFeatures);
+
+/// Removes a Feature from the FeatureModel
+///
+/// \param FM
+/// \param FeatureToBeDeleted
+void removeFeature(FeatureModel &FM, detail::FeatureVariantTy FeatureToBeDeleted,
+                   bool Recursive = false);
 
 /// Set commit of a FeatureModel.
 ///
--- a/include/vara/Feature/FeatureModelTransaction.h
+++ b/include/vara/Feature/FeatureModelTransaction.h
@@ -990,7 +990,7 @@ void addFeature(FeatureModel &FM, std::unique_ptr<Feature> NewFeature,
 /// \param NewFeatures
 void addFeatures(
     FeatureModel &FM,
-    std::vector<std::tuple<std::unique_ptr<Feature>, Feature *>> NewFeatures);
+    std::vector<std::pair<std::unique_ptr<Feature>, Feature *>> NewFeatures);
 
 /// Removes a Feature from the FeatureModel
 ///
--- a/include/vara/Feature/FeatureModelTransaction.h
+++ b/include/vara/Feature/FeatureModelTransaction.h
@@ -996,7 +996,8 @@ void addFeatures(
 ///
 /// \param FM
 /// \param FeatureToBeDeleted
-void removeFeature(FeatureModel &FM, detail::FeatureVariantTy FeatureToBeDeleted,
+void removeFeature(FeatureModel &FM,
+                   detail::FeatureVariantTy FeatureToBeDeleted,
                    bool Recursive = false);
 
 /// Set commit of a FeatureModel.
--- a/lib/Feature/FeatureModelTransaction.cpp
+++ b/lib/Feature/FeatureModelTransaction.cpp
@@ -18,10 +18,11 @@ void addFeature(FeatureModel &FM, std::unique_ptr<Feature> NewFeature,
 
 void addFeatures(
     FeatureModel &FM,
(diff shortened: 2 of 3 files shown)


include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -980,6 +980,26 @@ class FeatureModelModifyTransactionBase {
...
+///
+/// The vector of Features contains tuples of a parent and a new Feature.
+/// If a Parent is passed it needs to be already in the FeatureModel,
+/// otherwise, root is assumed as the parent Feature.
+///
+/// \param FM
+/// \param NewFeatures
+void addFeatures(
+    FeatureModel &FM,
+    std::vector<std::tuple<std::unique_ptr<Feature>, Feature *>> NewFeatures);
```
vulder: We only have two elements, so we could just use `std::pair` here, which is a special case of tuple.

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -16,6 +16,24 @@ void addFeature(FeatureModel &FM, std::unique_ptr<Feature> NewFeature,
   Trans.commit();
 }
 
+void addFeatures(
+    FeatureModel &FM,
+    std::vector<std::tuple<std::unique_ptr<Feature>, Feature *>> NewFeatures) {
+  auto Trans = FeatureModelModifyTransaction::openTransaction(FM);
+  for (auto &Tuple : NewFeatures) {
```
vulder: The loop var could have a better name, like `NewFeature`

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -832,4 +832,81 @@ TEST_F(FeatureModelRemoveFeatureTransactionTest,
...
+};
+
+TEST_F(FeatureModelTransactionTest, addFeatureToModel) {
+  size_t FMSizeBefore = FM->size();
+
+  vara::feature::addFeature(*FM,
+                            std::move(std::make_unique<BinaryFeature>("ab")),
+                            FM->getFeature("a")); // committed automatically
+
+  // Changes should be visible on the new model
```
vulder: This comment is confusing, why do we have a new model here?

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -832,4 +832,81 @@ TEST_F(FeatureModelRemoveFeatureTransactionTest,
...
+  std::vector<std::tuple<std::unique_ptr<Feature>, Feature *>> NewFeatures;
+  NewFeatures.emplace_back(std::make_tuple(
+      std::make_unique<BinaryFeature>("ab"), FM->getFeature("a")));
+  NewFeatures.emplace_back(std::make_tuple(
+      std::make_unique<BinaryFeature>("ac"), FM->getFeature("a")));
+
+  vara::feature::addFeatures(*FM,
+                             std::move(NewFeatures)); // committed automatically
+
+  // Changes should be visible on the new model
```
vulder: same

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -832,4 +832,81 @@ TEST_F(FeatureModelRemoveFeatureTransactionTest,
...
+  std::vector<std::tuple<std::unique_ptr<Feature>, Feature *>> NewFeatures;
+  NewFeatures.emplace_back(std::make_tuple(
+      std::make_unique<BinaryFeature>("ab"), FM->getFeature("a")));
+  NewFeatures.emplace_back(std::make_tuple(
+      std::make_unique<BinaryFeature>("ac"), FM->getFeature("a")));
+  vara::feature::addFeatures(*FM,
+                             std::move(NewFeatures)); // committed automatically
+
+  // Remove one Feature
+  // std::variant<std::string, Feature *>
```
vulder: ?

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -832,4 +832,81 @@ TEST_F(FeatureModelRemoveFeatureTransactionTest,
...
+  vara::feature::removeFeature(*FM, FM->getFeature("ab"));
+
+  // Changes should be visible on the new model
```
vulder: same as above

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -832,4 +832,81 @@ TEST_F(FeatureModelRemoveFeatureTransactionTest,
...
+  }
+
+  std::unique_ptr<FeatureModel> FM;
+};
+
+TEST_F(FeatureModelTransactionTest, addFeatureToModel) {
+  size_t FMSizeBefore = FM->size();
+
+  vara::feature::addFeature(*FM,
+                            std::move(std::make_unique<BinaryFeature>("ab")),
```
```suggestion
                            std::make_unique<BinaryFeature>("ab"),
```
`move` is not required here because the created unique_ptr is already an rvalue

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -832,4 +832,81 @@ TEST_F(FeatureModelRemoveFeatureTransactionTest,
...
+  // std::variant<std::string, Feature *>
+  vara::feature::removeFeature(*FM, FM->getFeature("ab"));
+
+  // Changes should be visible on the new model
+  EXPECT_EQ(FMSizeBefore, FM->size() - 1);
+  EXPECT_TRUE(FM->getFeature("a"));
+  EXPECT_TRUE(llvm::isa<RootFeature>(FM->getFeature("a")->getParentFeature()));
+  EXPECT_TRUE(FM->getFeature("ac")); // Change should be visible
+  EXPECT_EQ(FM->getFeature("a"), FM->getFeature("ac")->getParentFeature());
+}
```
vulder: This test does not test what it wants to test. We want to guarantee that `"ab"` is remove after being added. But we never check for `"ab"`. If we remove the implementation of `removeFeature` this test still passes.

vulder (changes requested):

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -16,6 +16,25 @@ void addFeature(FeatureModel &FM, std::unique_ptr<Feature> NewFeature,
   Trans.commit();
 }
 
+void addFeatures(
+    FeatureModel &FM,
+    std::vector<std::pair<std::unique_ptr<Feature>, Feature *>> NewFeatures) {
+  auto Trans = FeatureModelModifyTransaction::openTransaction(FM);
+  for (auto &NewFeature : NewFeatures) {
+    Trans.addFeature(std::move(std::get<0>(NewFeature)),
+                     std::get<1>(NewFeature));
```
```suggestion
    Trans.addFeature(std::move(NewFeature.first),
                     NewFeature.second);
```

vulder (approved):

padupr (approved):

PR #52, opened 2021-05-16
Implements removeFeatures for FeatureModelTransaction

MiriamMarx:


MiriamMarx:
How should the following scenario be handled (see test removeFeaturesFromModelNotPossible):
We have recursive=false and want to delete a Feature that has children. Currently, I delete as much as possible of the provided list and exit the function. Should I rather provide a vector of Features that could not be deleted? Or should I not delete anything at all if one Feature cannot be deleted?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -1000,6 +1000,23 @@ void removeFeature(FeatureModel &FM,
...
+                    bool Recursive = false);
+
+/// Removes multiple Features from the FeatureModel recursively
+///
+/// \param FM
+/// \param FeaturesToBeDeleted
+void removeFeaturesRecursively(
+    FeatureModel &FM,
+    std::vector<detail::FeatureVariantTy> FeaturesToBeDeleted);
+
```
vulder: Why do we want/need two different public functions when the upper one already has a recursive flag parameter?

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -34,6 +36,67 @@ void removeFeature(FeatureModel &FM,
...
+    // We are not in recursive mode and have only parents left with children
+    // that should not be deleted --> abort execution
+    if (DeleteFeatures.empty()) {
+      Trans.commit();
+      if (Recursive) {
+        removeFeaturesRecursively(FM, std::move(RemainingFeatures));
+      }
+      // TODO: return list of non-deletable Features?
+      return;
+    }
```
```suggestion
    }

```

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -34,6 +36,67 @@ void removeFeature(FeatureModel &FM,
...
+                    std::vector<detail::FeatureVariantTy> FeaturesToBeDeleted,
+                    bool Recursive) {
+  // Iterate over Features and deleted them iff they are leaves (don't have
+  // children).
+  // Even if we are in the recursive mode, we need to delete the children among
+  // FeaturesToBeDeleted first. Otherwise, the commit gets stuck in an endless
+  // loop.
+  std::vector<detail::FeatureVariantTy> RemainingFeatures;
+  std::copy(FeaturesToBeDeleted.begin(), FeaturesToBeDeleted.end(),
+            std::back_inserter(RemainingFeatures));
```
vulder: I don't think we need a copy here, we can just use the FeaturesToBeDeleted vector and a iterator.
We can use partition, to partition the FeatureToBeDeleted into others and leaves. Then work on the leafs partition. Then in the next cycle, repartition again to the the new leaves and repeat until end.
Afterwards, we hand the not deleted range, i.e., the left over partition that never become leaves to the clean up code.

If you want more infos, we can discuss this in persion :smile:

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -904,4 +904,128 @@ TEST_F(FeatureModelTransactionTest, removeFeatureFromModel) {
...
+
+  // Prepare Model with several Features
+  vara::feature::addFeature(*FM, std::make_unique<BinaryFeature>("ab"),
+                            FM->getFeature("a"));
+  vara::feature::addFeature(*FM, std::make_unique<BinaryFeature>("ac"),
+                            FM->getFeature("a"));
+  vara::feature::addFeature(*FM, std::make_unique<BinaryFeature>("ad"),
+                            FM->getFeature("ab"));
+  vara::feature::addFeature(*FM, std::make_unique<BinaryFeature>("ae"),
+                            FM->getFeature("ad"));
```
vulder, line 955: There is a new awesome function to add multiple features at one :wink:

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -904,4 +904,128 @@ TEST_F(FeatureModelTransactionTest, removeFeatureFromModel) {
...
+
+  // Prepare Features for deletion
+  std::vector<detail::FeatureVariantTy> FeaturesToBeDeleted;
+  FeaturesToBeDeleted.emplace_back(FM->getFeature("ad"));
+  FeaturesToBeDeleted.emplace_back(FM->getFeature("ac"));
+  FeaturesToBeDeleted.emplace_back(FM->getFeature("ae"));
```
```suggestion
  std::vector<detail::FeatureVariantTy> FeaturesToBeDeleted{
  FM->getFeature("ac"), FM->getFeature("ad"), FM->getFeature("ae")
  };
```
This should also work. (- formatting)

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -904,4 +904,128 @@ TEST_F(FeatureModelTransactionTest, removeFeatureFromModel) {
...
+
+  // Prepare Model with several Features
+  vara::feature::addFeature(*FM, std::make_unique<BinaryFeature>("ab"),
+                            FM->getFeature("a"));
+  vara::feature::addFeature(*FM, std::make_unique<BinaryFeature>("ac"),
+                            FM->getFeature("a"));
+  vara::feature::addFeature(*FM, std::make_unique<BinaryFeature>("ad"),
+                            FM->getFeature("ab"));
+  vara::feature::addFeature(*FM, std::make_unique<BinaryFeature>("ae"),
+                            FM->getFeature("ad"));
```
vulder, line 987: most of the previous feedback also applies to the following tests

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -904,4 +904,128 @@ TEST_F(FeatureModelTransactionTest, removeFeatureFromModel) {
...
+  FeaturesToBeDeleted.emplace_back(FM->getFeature("ad"));
+  FeaturesToBeDeleted.emplace_back(FM->getFeature("ac"));
+  FeaturesToBeDeleted.emplace_back(FM->getFeature("ae"));
+
+  vara::feature::removeFeatures(*FM, FeaturesToBeDeleted);
+
+  EXPECT_EQ(FMSizeBefore, FM->size() - 1);
+  EXPECT_TRUE(FM->getFeature("a"));
+  EXPECT_TRUE(llvm::isa<RootFeature>(FM->getFeature("a")->getParentFeature()));
+  EXPECT_TRUE(FM->getFeature("ab")); // Change should be visible
```
vulder: please remove comments that are not ment for this test but a copy paste from another

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -904,4 +904,128 @@ TEST_F(FeatureModelTransactionTest, removeFeatureFromModel) {
...
+
+  // Prepare Features for deletion
+  std::vector<detail::FeatureVariantTy> FeaturesToBeDeleted;
+  FeaturesToBeDeleted.emplace_back(FM->getFeature("ad"));
+  FeaturesToBeDeleted.emplace_back(FM->getFeature("ac"));
+  FeaturesToBeDeleted.emplace_back(FM->getFeature("ae"));
+
+  vara::feature::removeFeatures(*FM, FeaturesToBeDeleted);
+
+  EXPECT_EQ(FMSizeBefore, FM->size() - 1);
```
vulder: Wouldn't it be easier to understand to "remember" the FMSizeBefore after the new features where added. Currently, one has to understand that 4 are added and 3 are removed and then there is one less afterwards?
This seems confusing.

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -904,4 +904,128 @@ TEST_F(FeatureModelTransactionTest, removeFeatureFromModel) {
...
+  vara::feature::removeFeatures(*FM, FeaturesToBeDeleted, true);
+
+  EXPECT_EQ(FMSizeBefore, FM->size() - 1);
+  EXPECT_TRUE(FM->getFeature("a"));
+  EXPECT_TRUE(llvm::isa<RootFeature>(FM->getFeature("a")->getParentFeature()));
+  EXPECT_TRUE(FM->getFeature("ab")); // Change should be visible
+  EXPECT_EQ(FM->getFeature("a"), FM->getFeature("ab")->getParentFeature());
+  EXPECT_FALSE(FM->getFeature("ac"));
+  EXPECT_FALSE(FM->getFeature("ad"));
+  EXPECT_FALSE(FM->getFeature("ae"));
```
vulder: these should have comments that mention why they are remove like:
```suggestion
  EXPECT_FALSE(FM->getFeature("ac"));
  EXPECT_FALSE(FM->getFeature("ad"));
  // should be remove in recursive mode because ae is a child of ad
  EXPECT_FALSE(FM->getFeature("ae"));
```

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -904,4 +904,128 @@ TEST_F(FeatureModelTransactionTest, removeFeatureFromModel) {
...
+}
+
+TEST_F(FeatureModelTransactionTest, removeFeaturesFromModelNotPossible) {
```
vulder, line 978: This needs a comment, the name is not enough to understand the context.
Why is it not possible to be remove? Or why should it not be possible to remove something?

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -904,4 +904,128 @@ TEST_F(FeatureModelTransactionTest, removeFeatureFromModel) {
...
+
+  vara::feature::removeFeatures(*FM, FeaturesToBeDeleted);
+
+  EXPECT_EQ(FMSizeBefore, FM->size() - 3);
+  EXPECT_TRUE(FM->getFeature("a"));
+  EXPECT_TRUE(llvm::isa<RootFeature>(FM->getFeature("a")->getParentFeature()));
+  EXPECT_TRUE(FM->getFeature("ab")); // Change should be visible
+  EXPECT_EQ(FM->getFeature("a"), FM->getFeature("ab")->getParentFeature());
+  EXPECT_FALSE(FM->getFeature("ac"));
+  EXPECT_TRUE(FM->getFeature("ad")); // only true, because could not be deleted
```
```suggestion
  // ad should not be removed in without recursive mode because it this has a child
  EXPECT_TRUE(FM->getFeature("ad"));
```
Please try to use correct grammar

vulder (changes requested):

include/vara/Feature/FeatureModelTransaction.h
MiriamMarx: Actually, I would like to have the function as private, but there is no way in namespaces, right? Same counts for countNumberOfSuccessors(...) that I just created. Both are just own functions for code-readability. 
Any suggestions how to handle them?

unittests/Feature/FeatureModelTransaction.cpp
MiriamMarx: Yes, but since it does a commit in the very end, I cannot use it to create a tree-like hierarchy of my features. It would work out if all Features are just children of a.

unittests/Feature/FeatureModelTransaction.cpp
MiriamMarx: Done

unittests/Feature/FeatureModelTransaction.cpp
MiriamMarx: Done

lib/Feature/FeatureModelTransaction.cpp
MiriamMarx: Yes, let's discuss this :-D

include/vara/Feature/FeatureModelTransaction.h
vulder: Don't put them (the declaration) in the public header (include/vara/feature/*), but make it only local to the cpp file (can also be split into a declaration on the top and an implementation at the bottom) or create a private header file in the source dir (lib/Feature/*).

unittests/Feature/FeatureModelTransaction.cpp
vulder: Oh, hmm then maybe we need to adapt the impl of addFeatures. My inital thoughts were that this kind of scenario should be possible. So, maybe we need add them all an after wards commit. The only thing we need to make sure there is that parents are added before children because otherwise the add modification would not work.

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -50,46 +52,82 @@ void removeFeatures(FeatureModel &FM,
...
                          return binary_search(DeleteFeatures.begin(),
-                                              DeleteFeatures.end(), x);
+                                              DeleteFeatures.end(), X);
                        }),
         RemainingFeatures.end());
     Trans.commit();
   }
 }
 
+int countNumberOfSuccessors(FeatureTreeNode *F) {
```
vulder: maybe we should integrate this into the FeatureTreeNode class or at least on the corresponding header

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -93,6 +93,10 @@ class FeatureTreeNode : private llvm::DGNode<FeatureTreeNode, FeatureTreeNode> {
     return FS;
   }
 
+  /// Checks if a FeatureTreeNode has children or is a leave.
+  /// \return true, if the FeatureTreeNode does not have children/is a leave.
```
vulder:
```suggestion
  /// Checks if a FeatureTreeNode has children or is a leave.
  ///
  /// \return true, if the FeatureTreeNode does not have children, i.e., is a leave.
```

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -1,6 +1,10 @@
 #include "vara/Feature/FeatureModelTransaction.h"
 
+#include "vara/Utils/VariantUtil.h"
```
```suggestion
#include "vara/Utils/VariantUtil.h"
```

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -36,6 +40,86 @@ void removeFeature(FeatureModel &FM,
...
 
+bool fvIsLeave(FeatureModel &FM, detail::FeatureVariantTy &FV) {
+  Feature *ActualFeature;
+  std::visit(
+      Overloaded{[&ActualFeature, &FM](Feature *F) { ActualFeature = F; },
+                 [&ActualFeature, &FM](string &F) {
+                   ActualFeature = FM.getFeature(F);
+                 }},
+      FV);
+  return ActualFeature->isLeave();
```
vulder: What happens here if the FM does not have a feature with the given name?

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -36,6 +40,86 @@ void removeFeature(FeatureModel &FM,
   Trans.commit();
 }
 
+bool fvIsLeave(FeatureModel &FM, detail::FeatureVariantTy &FV) {
+  Feature *ActualFeature;
+  std::visit(
+      Overloaded{[&ActualFeature, &FM](Feature *F) { ActualFeature = F; },
+                 [&ActualFeature, &FM](string &F) {
+                   ActualFeature = FM.getFeature(F);
```
```suggestion
                 [&ActualFeature, &FM](string &FName) {
                   ActualFeature = FM.getFeature(FName);
```

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -36,6 +40,86 @@ void removeFeature(FeatureModel &FM,
...
+  }
+  return false;
+}*/
+
+void removeFeatures(FeatureModel &FM,
+                    std::vector<detail::FeatureVariantTy> FeaturesToBeDeleted,
+                    bool Recursive) {
+  removeFeatures(FM, FeaturesToBeDeleted.begin(), FeaturesToBeDeleted.end(),
+                 Recursive);
+}
```
vulder: This impl is quite short, so it could be moved to the header and merged with the declaration.

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -36,6 +40,86 @@ void removeFeature(FeatureModel &FM,
...
+  if (NoDeleteIt == End) {
+    // TODO: return list of non-deletable Features?
+    return;
+  }
+
+  // remove leaves
+  auto Trans = FeatureModelModifyTransaction::openTransaction(FM);
+  auto it = NoDeleteIt;
+  while (it != End) {
+    Trans.removeFeature(*it, true);
```
vulder: Should we not use `Recursive` here? When we pass `true` we always delete them recursively. On the other hand, we only remove leaves, so I guess, it does not matter.

vulder (changes requested):
@s9latimm deleting features has a few edge cases that we should talk about, especially, how we handle removal of features with children and multiple removals in one commit that "in theory" are ok but currently fail because of modification checks. Do you have time tomorrow to talk this over?

lib/Feature/FeatureModelTransaction.cpp
MiriamMarx: Done

lib/Feature/FeatureModelTransaction.cpp
MiriamMarx: Done

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -93,6 +93,11 @@ class FeatureTreeNode : private llvm::DGNode<FeatureTreeNode, FeatureTreeNode> {
     return FS;
   }
 
+  /// Checks if a FeatureTreeNode has children or is a leave.
+  ///
+  /// \return true, if the FeatureTreeNode does not have children, i.e., is a leave.
```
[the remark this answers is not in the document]
MiriamMarx: Done

lib/Feature/FeatureModelTransaction.cpp
MiriamMarx: Done --> Such elements won't be deleted.

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -36,6 +39,98 @@ void removeFeature(FeatureModel &FM,
...
+  Feature *ActualFeature = getActualFeature(FM, FV);
+  // if Feature does not exist in FM
+  if (ActualFeature == nullptr) {
+    return std::nullopt;
+  }
+  return std::optional<bool>{ActualFeature->isLeave()};
+}
+
+std::optional<bool>
+canBeDeletedRecursively(FeatureModel &FM, detail::FeatureVariantTy &FV,
```
vulder, line 71: Could you please write a quick doc string for this function.

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -36,6 +39,98 @@ void removeFeature(FeatureModel &FM,
...
+                        const std::set<FeatureTreeNode *> &OtherFeatures) {
+  Feature *ActualFeature = getActualFeature(FM, FV);
+  // if Feature does not exist in FM
+  if (ActualFeature == nullptr) {
```
```suggestion
  if (ActualFeature == nullptr) {  // Feature does not exist in FM
```

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -1002,6 +1002,31 @@ void removeFeature(FeatureModel &FM,
                    detail::FeatureVariantTy FeatureToBeDeleted,
                    bool Recursive = false);
 
+/// Removes multiple Features from the FeatureModel
+///
+/// \param FM
+/// \param Begin
+/// \param End
+/// \param Recursive
```
vulder, line 1110: Return value not mentioned in docs

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -1002,6 +1002,31 @@ void removeFeature(FeatureModel &FM,
...
+removeFeatures(FeatureModel &FM,
+               std::vector<detail::FeatureVariantTy>::iterator Begin,
+               std::vector<detail::FeatureVariantTy>::iterator End,
+               bool Recursive = false);
+
+/// Removes multiple Features from the FeatureModel
+///
+/// \param FM
+/// \param FeaturesToBeDeleted
+/// \param Recursive
```
vulder, line 1122: Return value not mentioned in docs

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -36,6 +39,98 @@ void removeFeature(FeatureModel &FM,
...
+          return false;
+        }
+        std::set<FeatureTreeNode *> OtherFeatures;
+        std::transform(
+            Begin, End,
+            std::inserter(OtherFeatures, std::next(OtherFeatures.begin())),
+            [&FM, &OtherFeatures](detail::FeatureVariantTy &FV) {
+              return getActualFeature(FM, FV);
+            });
+        OtherFeatures.erase(nullptr);
```
```suggestion
        // Remove nullptr, in case, on of the specified Features could 
        // not be found in the FeatureModel 
        OtherFeatures.erase(nullptr);
```

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -36,6 +39,98 @@ void removeFeature(FeatureModel &FM,
...
+        return canBeDeletedRecursively(FM, FV, OtherFeatures).value_or(false);
+      });
+
+  // check if something can be deleted --> if not and return non-deletable
+  // features
+  if (DeleteIt == Begin) {
+    return std::vector(Begin, End);
+  }
+
+  // remove deletable features
```
```suggestion
  // perform actual removal of deletable features
```

vulder (changes requested):
Overall looks quite clean now.

unittests/Feature/FeatureModelTransaction.cpp
vulder: Can you please create an issue so we don't forget this in the future and like it with a comment to the corresponding locations. The format ist `TODO (se-sic/VaRA#???): short description`

include/vara/Feature/FeatureModelTransaction.h
MiriamMarx: Done

include/vara/Feature/FeatureModelTransaction.h
MiriamMarx: Done

lib/Feature/FeatureModelTransaction.cpp
MiriamMarx: Done

unittests/Feature/FeatureModelTransaction.cpp
MiriamMarx: Done

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -93,6 +93,12 @@ class FeatureTreeNode : private llvm::DGNode<FeatureTreeNode, FeatureTreeNode> {
     return FS;
   }
 
+  /// Checks if a FeatureTreeNode has children or is a leave.
```
s9latimm:
```suggestion
  /// Checks if a FeatureTreeNode has children or is a leaf.
```

include/vara/Feature/FeatureTreeNode.h
```diff
@@ -93,6 +93,12 @@ class FeatureTreeNode : private llvm::DGNode<FeatureTreeNode, FeatureTreeNode> {
...
+  ///
+  /// \return true, if the FeatureTreeNode does not have children, i.e., is a
+  /// leave.
+  bool isLeave() { return this->children().empty(); }
```
s9latimm:
```suggestion
  bool isLeaf() { return this->children().empty(); }
```

s9latimm (approved):
LGTM

vulder (approved):

vulder:
related se-sic/VaRA#749

PR #54, opened 2021-05-24
Fixes to Merging algorithm

padupr:
Last minute fixes.


vulder:
@padupr What's the current status of this PR?

include/vara/Feature/FeatureModel.h
```diff
@@ -286,6 +286,14 @@ class FeatureModel {
     return Constraints.back().get();
   }
 
+  void removeConstraint(Constraint *C) {
+    Constraints.erase(
+        std::find_if(Constraints.begin(), Constraints.end(),
+                     [C](const std::unique_ptr<Constraint> &UniC) {
+                       return UniC.get() == C;
```
s9latimm: Here we compare two pointers (instead of the underlying constraints) -- maybe we should replace this in the future with tree based constraint comparison.

lib/Feature/FeatureModelWriter.cpp
```diff
@@ -25,8 +25,9 @@ static constexpr char ENCODING[] = "UTF-8";
 int FeatureModelXmlWriter::writeFeatureModel(std::string Path) {
   int RC;
   std::unique_ptr<xmlTextWriter, void (*)(xmlTextWriterPtr)> Writer(
-      xmlNewTextWriterFilename(Path.data(), 0), &xmlFreeTextWriter);
+      xmlNewTextWriterFilename(Path.c_str(), 0), &xmlFreeTextWriter);
```
s9latimm: Does this change anything?

https://stackoverflow.com/questions/194634/string-c-str-vs-data

unittests/Feature/FeatureModelParser.cpp
```diff
@@ -29,17 +29,6 @@ TEST(FeatureModelParser, onlyChildren) {
...
-  auto FS = llvm::MemoryBuffer::getFileAsStream(
-      getTestResource("test_only_parents.xml"));
-  assert(FS);
-
-  auto P = FeatureModelXmlParser(FS.get()->getBuffer().str());
-
-  EXPECT_TRUE(P.verifyFeatureModel());
-  EXPECT_TRUE(P.buildFeatureModel());
-}
-
```
s9latimm: Why is this test disabled? The goal was to test solely the behavior of the parser.

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -147,9 +178,9 @@ mergeFeatureModels(FeatureModel &FM1, FeatureModel &FM2, bool Strict) {
   if (F1.getKind() == Feature::FeatureKind::FK_ROOT) {
     return true;
   }
-  if (F1.getParent()->getKind() != F2.getParent()->getKind()) {
-    return false;
-  }
+  //  if (F1.getParent()->getKind() != F2.getParent()->getKind()) {
+  //    return false;
+  //  }
```
s9latimm: Why is this commented out?

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -394,6 +397,228 @@ TEST_F(FeatureModelMergeTransactionTest, AcceptNonStrictDifferenceKind) {
...
+  auto FMMerged = mergeFeatureModels(*FM2, *FM1);
+  ASSERT_TRUE(FMMerged);
+
+  ASSERT_TRUE(FMMerged->getFeature("a"));
+  ASSERT_TRUE(FMMerged->getFeature("aa"));
+  EXPECT_EQ(FMMerged->getFeature("aa")->getParent()->getKind(),
+            FeatureTreeNode::NodeKind::NK_RELATIONSHIP);
+  ASSERT_TRUE(FMMerged->getFeature("ab"));
+  EXPECT_EQ(FMMerged->getFeature("ab")->getParent()->getKind(),
+            FeatureTreeNode::NodeKind::NK_RELATIONSHIP);
```
s9latimm, line 481: This may be refactored with using `TearDown` in a base test.

s9latimm (changes requested):
Over all LGTM, just some remarks / questions unrelated to the actual algorithm.

unittests/Feature/FeatureModelTransaction.cpp
```diff
@@ -1,9 +1,12 @@
 #include "vara/Feature/FeatureModelTransaction.h"
 #include "vara/Feature/FeatureModelBuilder.h"
 
+#include "UnittestHelper.h"
 #include "gtest/gtest.h"
 
+#include <llvm/Support/MemoryBuffer.h>
 #include <memory>
+#include <vara/Feature/FeatureModelParser.h>
```
```suggestion
#include "vara/Feature/FeatureModelTransaction.h"
#include "vara/Feature/FeatureModelBuilder.h"
#include "vara/Feature/FeatureModelParser.h"

#include "UnittestHelper.h"

#include "gtest/gtest.h"

#include "llvm/Support/MemoryBuffer.h"

#include <memory>

```

unittests/Feature/FeatureModel.cpp
```diff
@@ -224,6 +224,34 @@ TEST_F(FeatureModelTest, dfs) {
 //                    FeatureModelConsistencyChecker Tests
```
vulder: Copy paste error, I think you wanted to put the test name there.

lib/Feature/FeatureModelTransaction.cpp
```diff
@@ -7,10 +7,15 @@ namespace vara::feature {
 bool mergeSubtree(FeatureModelCopyTransaction &Trans, FeatureModel const &FM,
                   Feature &F, bool Strict);
 
+bool mergeSubtree(FeatureModelCopyTransaction &Trans, FeatureModel const &FM,
+                  Relationship &R, bool Strict);
```
vulder: I'm not 100% sure what this method does, but maybe the name is just confusing.
Does this merge feature subtree that are related to the Relationship or does it merge relationships?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -803,14 +861,19 @@ class FeatureModelCopyTransactionBase {
   }
 
   Constraint *addConstraintImpl(std::unique_ptr<Constraint> NewConstraint) {
-    if (!FM) {
-      return nullptr;
-    }
+    assert(FM && "FeatureModel is null.");
```
vulder: Do we want to just assert here or return by default a nullptr when we are called with no FM?
@s9latimm What do you think?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -365,6 +414,15 @@ class RemoveFeatureFromModel : public FeatureModelModification {
       }
     }
 
+    while (!F->constraints().empty()) {
```
vulder, line 434: If I understand this correctly, we go to the most upper constraint and delete it. You can we prevent that this delete much more than was intended? Not sure if a edge case can happen here, where we remove to much.

include/vara/Feature/FeatureModel.h
vulder: If we don't have them yet, a TODO that refers to the issue would be good.

vulder (changes requested):

include/vara/Feature/FeatureModel.h
s9latimm: https://github.com/se-sic/VaRA/issues/701

include/vara/Feature/FeatureModelTransaction.h
padupr: I had a look at other Transactions. It's handled in three ways: check and return nullptr, assert, no check. I think this should be solved after error codes are implemented.

include/vara/Feature/FeatureModelTransaction.h
padupr: This code should: 

1. Pick a constraint of the feature
2. Traverse upwards to get the constraint's expression tree root
3. Remove the whole Constraint from the model, this will also remove it from all other related features
4. Continue with the next constraint

We might want to offer a choice on the behavior in the future, e.g. error if constraints are present, try to resolve the deletion based on rules, remove all present constraints on the feature entirely (current solution).

lib/Feature/FeatureModelTransaction.cpp
padupr: It's unused so I removed it. Originally, planned to overload the function to handle Relationships differently during merging.

lib/Feature/FeatureModelTransaction.cpp
padupr: I needed to allow merging of subtrees if the relationship is not present in both, so the parent node is no longer required to be equal.

unittests/Feature/FeatureModelParser.cpp
padupr: Not sure why I removed that. Reverted.

lib/Feature/FeatureModelWriter.cpp
vulder: should not matter anymore

lib/Feature/FeatureModelTransaction.cpp
vulder: So can we remove this or do we need a TODO to implement something else?

lib/Feature/FeatureModelTransaction.cpp
vulder: So, if you don't need that anymore a updated name should suffice

include/vara/Feature/FeatureModelTransaction.h
vulder: Yes, error codes should make this case clear, so let's keep it that way.

include/vara/Feature/FeatureModelTransaction.h
vulder: Ok, sounds ok to me. Could you please create an issue for that an reference it here with an TODO, so we come back to this later :)

vulder:
What is the current state of this PR? @padupr could you please, if you find time, go over the open remarks and close them if they are resolved.

lib/Feature/FeatureModelTransaction.cpp
padupr: Removed it, no need to implement something else.

vulder:
@padupr I pulled in the current fixes on dev. Otherwise, we need to remove the issues reported from the CI.

vulder:
@s9latimm I updated the PR to our ne dev state. Can you please double check this PR, especially the TODO part regarding error handling. What should we do in this case? Do we need a specific error here?

s9latimm (approved):

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -326,6 +342,46 @@ class AddFeatureToModel : public FeatureModelModification {
...
+    if (auto E = (*this)(FM); !E) {
+      return E.getError();
+    }
+    return Ok();
+  }
+
+  Result<FTErrorCode> operator()(FeatureModel &FM) {
+    // TODO: Any error handling needed here?
+    assert(RemoveConstraint.getParent() == nullptr &&
+           "Subtree deletion is not supported");
```
s9latimm: We should talk about this today -- as far as I understand, uncoupling currently only works for this case.

vulder (approved):

s9latimm (approved):

vulder:
- [x] Depends on #133

vulder:
@boehmseb @Kallistos please have a look. I tried to rebase and clean up the work from Pascal but an additional feedback would be good.

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -254,6 +258,14 @@ class FeatureModelModification {
     return FM.addConstraint(std::move(Constraint));
   }
 
+  static void removeConstraint(FeatureModel &FM, Constraint *R) {
```
boehmseb: Why `Consraint *R` here and `Constraint *C` below?
Do they have different meaning, or should they be named the same?

include/vara/Feature/FeatureModel.h
```diff
@@ -399,6 +399,48 @@ class FeatureModel {
...
+    }
+
+    { // Handle mixed constraints
+      auto MCIt =
+          std::find_if(MixedConstraints.begin(), MixedConstraints.end(),
+                       [C](const std::unique_ptr<MixedConstraint> &UniC) {
+                         return UniC->constraint() == C;
+                       });
+      if (MCIt != MixedConstraints.end()) {
+
```
boehmseb:
```suggestion
```

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -310,8 +322,12 @@ class AddFeatureToModel : public FeatureModelModification {
       return ALREADY_PRESENT;
     }
     if (Parent) {
-      setParent(*InsertedFeature, *Parent);
-      addEdge(*Parent, *InsertedFeature);
+      FeatureTreeNode *ParentNode = Parent;
+      if (!Parent->getChildren<Relationship>(1).empty()) {
+        ParentNode = *(Parent->getChildren<Relationship>(1).begin());
+      }
```
boehmseb, line 328: Just to make sure I understand this correctly:
This is required because we model (xor/or) relationships as separate nodes so the relationship node has to be set as parent if one exists, correct?

This assumes that a Feature can have exactly one (or no) relationship as child and if it has, there are no features as direct children allowed. Is this assumption validated anywhere?
I'm asking because this might become an issue when integrating UVL because UVL does not have these constraints, i.e., a feature can have multiple types of relationships as children at once (I'm not sure how about multiple instances of the same type, e.g., multiple or groups).

boehmseb (approved):

include/vara/Feature/FeatureModelTransaction.h
vulder: I'm not sure, if it's currently not possible to have multiple ones, we probably can adapt the API to make it viable.

include/vara/Feature/FeatureModelTransaction.h
vulder: Don't know, probably an over site from @padupr , I'll rename this to C.

PR #60, opened 2021-07-17
Fix XML parsing and move test resources to own folder

s9latimm:
related se-sic/VaRA/issues/770


lib/Feature/FeatureModelParser.cpp
```diff
@@ -15,6 +15,10 @@ using std::make_unique;
 
 namespace vara::feature {
 
+std::string trim(const std::string &S) {
+  return llvm::StringRef(S).trim().str();
```
```suggestion
std::string trim(llvm::StringRef S) {
  return S.trim().str();
```
Would it not be simpler to directly take a string ref if we, in call cases convert it to one anyhow?

unittests/resources/xml/test_trim.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vulder: Maybe we should name the file `test_fm_with_whitespaces.xml` with just trim it's not really clear why this file is special.

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -682,7 +682,9 @@ class AddChild : public FeatureModelModification {
                          Parent);
     assert(C && P);
 
-    removeEdge(*C->getParent(), *C);
+    if (C->getParent()) {
+      removeEdge(*C->getParent(), *C);
+    }
```
vulder, line 687: Is this a special case where we remove something from root? (so: `C == Root`)?

vulder (approved):

include/vara/Feature/FeatureModelTransaction.h
s9latimm: Yes, exactly that is the intention. Additionally we are now more robust regarding removal of parents beforehand, as they are reassigned in this step anyway.

vulder (approved):

PR #61, opened 2021-07-17
Add Transaction Errors

s9latimm:
Add errors for transactions and handle them in parser / builder.

related https://github.com/se-sic/VaRA/issues/770
related https://github.com/se-sic/VaRA/issues/723

closes https://github.com/se-sic/VaRA/issues/744


vulder (commented):
from  the overall idea this looks fine
for debugging, maybe asan helps, could be that one function is return void and we just try to forward the return value -> UB


*later we just need to switch the bool with an enum that encode the error conditions

include/vara/Feature/Error.h
```diff
@@ -0,0 +1,79 @@
...
+#include "llvm/Support/raw_ostream.h"
+
+#include <cassert>
+#include <iostream>
+#include <type_traits>
+#include <variant>
+
+namespace vara::feature {
+
+enum ErrorCode {
```
vulder:
```suggestion
enum FTErrorCode {
```
We should make it a bit clearer to what the ErrorCode actually belongs. So maybe `FTErrorCode` or `FeatureTransactionErrorCode` or `FeatureErrorCode`

include/vara/Feature/Error.h
```diff
@@ -0,0 +1,79 @@
...
+  MISSING_FEATURE,
+  MISSING_PARENT,
+  ALREADY_PRESENT,
+  MISSING_MODEL,
+  INCONSISTENT
+};
+
+template <typename ValueTy, typename ErrorTy = ErrorCode,
+          std::enable_if_t<!std::is_same_v<ValueTy, ErrorTy>, bool> = true>
+class ErrorOr {
```
vulder: Just use `llvm::ErrorOr`, we don't have to implement this here our selfs.
The only thing we need to do is make our errors std::error_code

include/vara/Feature/Error.h
```diff
@@ -0,0 +1,79 @@
...
+  std::variant<ValueTy, ErrorTy> Value;
+};
+
+class Result : public ErrorOr<std::monostate, ErrorCode> {
+public:
+  Result() : ErrorOr<std::monostate, ErrorCode>(std::monostate{}) {}
+  Result(ErrorCode Error) : ErrorOr<std::monostate, ErrorCode>(Error) {}
+
+  std::string toString() {
+    if (this->operator bool()) {
```
vulder:
```suggestion
    if (*this) {
```

include/vara/Feature/Error.h
```diff
@@ -0,0 +1,79 @@
...
+  [[nodiscard]] ErrorTy getError() {
+    assert(std::holds_alternative<ErrorTy>(Value));
+    return std::move(std::get<ErrorTy>(Value));
+  }
+
+protected:
+  std::variant<ValueTy, ErrorTy> Value;
+};
+
+class Result : public ErrorOr<std::monostate, ErrorCode> {
```
vulder: Hmm, this design looks a bit unnecessary complex. Why do we need a `Result` (+not a good name) type, when we just want to return an error code? An alternative could be to define the first state of the error code to `OK`, rename it to `****State` instead (or another name) of error and just always return that type.

Maybe, we could flip this on it's head and always return a `Foo<StatusCode, Value>` and have a special `Foo<StatusCode, void>` that represents a non return function. `Foo` then specifies that Value is only correct if `StatusCode` is `OK` (similar to an optional where we check if something is in there before we are allowed to access it).

include/vara/Feature/Result.h
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+private:
+  ErrorTy E;
+};
+
+} // namespace result
+
+template <typename ErrorTy, typename ValueTy = void>
+class Result {
```
vulder: Same the general `Result` imple looks like a utility to me, so Utils makes much more sense than putting it just in the Feature part.

include/vara/Feature/Result.h
```diff
@@ -0,0 +1,115 @@
...
+#include "vara/Utils/VariantUtil.h"
+
+#include <cassert>
+#include <iostream>
+#include <type_traits>
+#include <variant>
+
+namespace vara::feature {
+
+namespace result {
```
vulder: What's the idea behind putting this into it's own namespace?

include/vara/Feature/Result.h
```diff
@@ -0,0 +1,115 @@
...
+
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &OS,
+                                       const Ok<void> &Ok) {
+    OS << "OK";
+    return OS;
+  }
+};
+
+template <typename ErrorTy>
+class Error {
```
vulder: Why is the default `Error` class in the Result header? Shouldn't this be more in Utils?

include/vara/Feature/Result.h
```diff
@@ -0,0 +1,115 @@
...
+result::Ok<ValueTy> Ok(ValueTy V) {
+  return result::Ok<ValueTy>(std::move(V));
+}
+
+inline result::Ok<void> Ok() { return result::Ok<void>(); }
+
+template <typename ErrorTy>
+result::Error<ErrorTy> Error(ErrorTy E) {
+  return result::Error<ErrorTy>(std::move(E));
+}
```
vulder: This one looks nice (similar than rust :wink:).
But we should maybe additionally implement perfectly forwarding constructor calls with this pattern, so we can construct 'Foo' directly inside of the OK.

lib/Feature/FeatureModel.cpp
```diff
@@ -93,7 +93,9 @@ std::unique_ptr<FeatureModel> FeatureModel::clone() const {
     FMB.addConstraint(C->clone());
   }
 
-  return FMB.buildFeatureModel();
+  auto FM = FMB.buildFeatureModel();
+  assert(FM);
```
vulder, line 100: Maybe a quick message, what happened here.

include/vara/Feature/Result.h
```diff
@@ -0,0 +1,115 @@
+#ifndef VARA_FEATURE_RESULT_H
+#define VARA_FEATURE_RESULT_H
+
+#include "llvm/Support/raw_ostream.h"
+
+#include "vara/Utils/VariantUtil.h"
```
vulder:
```suggestion
#include "vara/Utils/VariantUtil.h"

#include "llvm/Support/raw_ostream.h"
```

include/vara/Feature/Result.h
```diff
@@ -0,0 +1,115 @@
...
+namespace vara::feature {
+
+namespace result {
+
+template <typename ValueTy>
+class Ok {
+public:
+  Ok(ValueTy V) : V(std::move(V)) {}
+
+  ValueTy operator*() { return std::move(V); }
```
vulder: I'm not sure if we should always move out the value, take a look at my proposal below, but also feel free to argu against it.

```suggestion
  ValueTy operator*() && { return extract_value(); }
  ValueTy operator*() & { return V; }
  
  ValueTy extract_value() { return std::move(V); }
```
In cases where we are an `rvalue` we can savely always move out the value, as it will not be needed afterwards.
If we are not, we "by default" copy the value out but offer and extra function `extract_value` to allow the caller to move out the value explicitly.

include/vara/Feature/Result.h
```diff
@@ -0,0 +1,115 @@
...
+    return OS;
+  }
+};
+
+template <typename ErrorTy>
+class Error {
+public:
+  Error(ErrorTy V) : E(std::move(V)) {}
+
+  ErrorTy operator*() { return std::move(E); }
```
vulder: Same here.

include/vara/Feature/Result.h
```diff
@@ -0,0 +1,115 @@
...
+};
+
+} // namespace result
+
+template <typename ErrorTy, typename ValueTy = void>
+class Result {
+public:
+  Result(result::Ok<ValueTy> V) : Variant(std::move(V)) {}
+
+  Result(result::Error<ErrorTy> E) : Variant(std::move(E)) {}
```
vulder: Probably needs the same forwarding ctors as I mentioned below.

include/vara/Feature/Result.h
```diff
@@ -0,0 +1,115 @@
...
+
+  /// Return false if there is an error.
+  operator bool() const {
+    return !std::holds_alternative<result::Error<ErrorTy>>(Variant);
+  }
+
+  [[nodiscard]] ValueTy operator*() {
```
vulder: Same case here for moving things out

include/vara/Feature/Error.h
```diff
@@ -0,0 +1,74 @@
...
+  ERROR,
+  ABORTED,
+  MISSING_FEATURE,
+  MISSING_PARENT,
+  ALREADY_PRESENT,
+  MISSING_MODEL,
+  INCONSISTENT
+};
+
+namespace result {
```
vulder: I don't thing this should be in another namespace, just `vara:feature` as the Spezialisation is directly related to features.

include/vara/Feature/Error.h
```diff
@@ -0,0 +1,74 @@
...
+  Error(FTErrorCode E) : E(E) {}
+
+  FTErrorCode operator*() { return E; }
+
+  operator bool() const { return false; }
+
+  friend llvm::raw_ostream &operator<<(llvm::raw_ostream &OS,
+                                       const Error<FTErrorCode> &Error) {
+    switch (Error.E) {
+      // TODO(s9latimm): meaningful error msg
```
vulder: Open  TODO

include/vara/Feature/FeatureModelParser.h
```diff
@@ -46,23 +46,23 @@ class FeatureModelXmlParser : public FeatureModelParser {
 public:
   explicit FeatureModelXmlParser(std::string Xml) : Xml(std::move(Xml)) {}
 
-  std::unique_ptr<FeatureModel> buildFeatureModel() override;
+  Result<FTErrorCode> verifyFeatureModel() override;
 
-  bool verifyFeatureModel() override;
+  std::unique_ptr<FeatureModel> buildFeatureModel() override;
```
vulder, line 54: Will we are at it, could you please add 1-2 sentences to comment the public API here. Now that we have docs, we should clean up missing comments as much as possible :laughing:

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -296,12 +295,17 @@ class AddFeatureToModel : public FeatureModelModification {
   friend class FeatureModelModification;
 
 public:
-  void exec(FeatureModel &FM) override { (*this)(FM); }
+  Result<FTErrorCode> exec(FeatureModel &FM) override {
+    if (auto E = (*this)(FM); !E) {
+      return Error(E.getError());
```
vulder: Can we not just return `E` here?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -341,13 +350,13 @@ class RemoveFeatureFromModel : public FeatureModelModification {
...
       // TODO (se-passau/VaRA#744): error, does not exist
-      return;
+      return Error(MISSING_FEATURE);
     }
 
     if (Recursive) {
       for (auto *C : F->getChildren<Feature>()) {
         // TODO (se-passau/VaRA#744): error, if call returns error (should not
-        // be possible)
+        //  be possible)
```
vulder: If this PR closes se-passau/VaRA#744 we should fix this here

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -341,13 +350,13 @@ class RemoveFeatureFromModel : public FeatureModelModification {
         GroupRoot);
     if (!F) {
       // TODO (se-passau/VaRA#744): error, does not exist
-      return;
+      return Error(MISSING_FEATURE);
```
vulder: I guess the comment above can be removed?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -356,9 +365,9 @@ class RemoveFeatureFromModel : public FeatureModelModification {
       }
     } else {
       if (!F->getChildren<Feature>().empty()) {
-        // TODO (se-passau/VaRA#744): error, non recursive 0remove on non leaf
-        // feature
-        return;
+        // TODO (se-passau/VaRA#744): error, non recursive remove on non leaf
+        //  feature
+        return Error(MISSING_FEATURE);
```
vulder: Is this correctly handled?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -442,7 +462,7 @@ class RemoveRelationshipFromModel : public FeatureModelModification {
         Parent);
     if (!P) {
       // TODO(se-passau/VaRA#744) handle this
-      return;
+      return Error(MISSING_PARENT);
```
vulder: Same here, should we remove the comment?

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -628,9 +690,18 @@ class SetRoot : public FeatureModelModification {
...
+    if (auto E = (*this)(FM); !E) {
+      return Error(E.getError());
+    }
+    return Ok();
+  }
 
-  RootFeature *operator()(FeatureModel &FM) {
+  Result<FTErrorCode, RootFeature *> operator()(FeatureModel &FM) {
+    if (FM.getRoot() && FM.getRoot()->getName() == Root->getName()) {
+      // TODO(s9latimm): swap root
```
vulder: Open TODO

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -856,26 +946,36 @@ class FeatureModelModifyTransactionBase {
...
+      if (auto E = ConsistencyCheck::isFeatureModelValid(*FM); !E) {
+        abortImpl();
+        return Error(E.getError());
+      }
+      FM = nullptr;
+      return Ok();
     }
-    return false;
-  };
+    abortImpl();
```
vulder, line 966:
```suggestion

    abortImpl();
```

vulder (changes requested):
The Result/Error/Ok design looks nice.

We should also add unittests for the new utility classes Result/Error/Ok, which don't depend on anything Feature.

include/vara/Utils/Result.h
```diff
@@ -0,0 +1,141 @@
...
+  ErrorTy E;
+};
+
+} // namespace result
+
+template <typename ErrorTy, typename ValueTy = void>
+class Result {
+public:
+  Result(result::Ok<ValueTy> V)
+      : Variant(std::forward<result::Ok<ValueTy>>(V)) {}
```
vulder: What is the idea between forward here?
I guess you miss understood what I meant with my comment. I did not mean to use std::forward here on the Ok value (that would need a move) but to allow the Ok/Error values to construct the carried value by passing the correct parameters.

If you like, we can talk about this next Monday or so.

s9latimm: Ah I see, probably we should talk about the remaining points anyway.

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -356,9 +364,9 @@ class RemoveFeatureFromModel : public FeatureModelModification {
       }
     } else {
       if (!F->getChildren<Feature>().empty()) {
-        // TODO (se-passau/VaRA#744): error, non recursive 0remove on non leaf
-        // feature
-        return;
+        // TODO (se-passau/VaRA#744): error, non recursive remove on non leaf
+        //  feature
```
vulder: Reference to the closed issue se-sic/VaRA#744

include/vara/Feature/FeatureModelTransaction.h
```diff
@@ -856,26 +944,37 @@ class FeatureModelModifyTransactionBase {
...
-          Modifications.begin(), Modifications.end(),
-          [this](const std::unique_ptr<FeatureModelModification> &FMM) {
-            FMM->exec(*FM);
-          });
-      if (ConsistencyCheck::isFeatureModelValid(*FM)) {
-        FM = nullptr;
-        return true;
+  Result<FTErrorCode> commitImpl() {
+    //    assert(FM && "Cannot commit Modifications without a FeatureModel
+    //    present.");
```
vulder: Out commented assert? Should this be included?

include/vara/Utils/Result.h
```diff
@@ -0,0 +1,153 @@
...
+template <typename ErrorTy, typename ValueTy = void>
+class Result {
+public:
+  Result(Ok<ValueTy> V) : Variant(std::move(V)) {}
+
+  Result(Error<ErrorTy> E) : Variant(std::move(E)) {}
+
+  template <typename OtherTy,
+            std::enable_if_t<!std::is_same_v<ValueTy, OtherTy>, bool> = true>
+  Result(Result<ErrorTy, OtherTy> Other) : Variant(Other.extract_error()) {}
```
vulder: What's the idea behind this impl? I would assume, if the error type is the same but the value type is different, we can take over the error. However, what I cannot figure out, what happens if `Other` does not contain an error, but a valid value of type `OtherTy`?

vulder (changes requested):

include/vara/Utils/Result.h
s9latimm: With the current way of constructing, either an assertion or exception of `std::get: wrong index for variant` will be thrown, if `Other` does not actually contain an `Error`.
I did not come up with a better idea on how to infer that beforehand yet.

include/vara/Feature/FeatureModelTransaction.h
s9latimm: This is not required anymore, as we already return an error if FM is not present, and may break with aborted transactions.

vulder (approved):

PR #62, opened 2021-07-18
Sxfm constraints

ChristianKaltenecker:
This is related to https://github.com/se-sic/VaRA/issues/773

Furthermore, I have fixed an issue in the syntax diagram of the constraints


ChristianKaltenecker:
The failed test depends on PR https://github.com/se-sic/vara-feature/pull/61.

include/vara/Feature/FeatureModelParser.h
```diff
@@ -225,6 +225,8 @@ class FeatureModelSxfmParser : public FeatureModelParser {
   std::string Sxfm;
   FeatureModelBuilder FMB;
   std::string Indentation = "\t";
+  std::map<std::string, std::string> IdentifierMap =
+      std::map<std::string, std::string>();
```
vulder:
```suggestion
  std::map<std::string, std::string> IdentifierMap;
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -540,6 +539,17 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlNode *FeatureTree) {
...
       }
 
+      // Search for identifier
+      std::regex Regex(R"(\([a-zA-Z\d_]*\))");
+      std::string CurrentLine = ToStringRef.str();
+      for (std::sregex_iterator Match = std::sregex_iterator(
+               CurrentLine.begin(), CurrentLine.end(), Regex);
+           Match != std::sregex_iterator(); Match++) {
+        std::string Identifier = (*Match).str();
+        Identifier = Identifier.substr(1, Identifier.size() - 2);
```
vulder, line 563: I get why you start at char 1 (to skip `:`) but why do you cut of 2 afterwards (1 for size() - 1, but where does the other character come from?)

lib/Feature/FeatureModelParser.cpp
```diff
@@ -540,6 +539,17 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlNode *FeatureTree) {
...
+        IdentifierMap[Identifier] = Name;
```
vulder, line 564:
```suggestion
        IdentifierMap[Identifier] = std::move(Name);
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -609,8 +619,64 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlNode *FeatureTree) {
...
+        << "Failed to read in constraints. Is the constraint field empty?\n";
+    return false;
+  }
+
+  std::stringstream Ss(reinterpret_cast<const char *>(
+      UniqueXmlChar(xmlNodeGetContent(Constraints), xmlFree).get()));
+  std::string To;
+
+  // Prepare the identifiers for later replacement
+  std::vector<std::string> Keys{};
```
```suggestion
  std::vector<std::string> Keys;
  Keys.reserve(IdentifierMap.size());
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -609,8 +619,64 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlNode *FeatureTree) {
...
+  for (const auto &[key, value] : IdentifierMap) {
+    auto It = Keys.begin();
+    Keys.insert(It, key);
+  }
```
```suggestion
  std::for_each(IdentifierMap.begin(), IdentifierMap.end(), [&Keys](const auto &Entry) { Keys.push_back(Entry.first); });
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -609,8 +619,64 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlNode *FeatureTree) {
...
+  std::sort(Keys.begin(), Keys.end(),
+            [](const std::string &First, const std::string &Second) {
+              return First.size() > Second.size();
+            });
```
```suggestion
            std::greater<std::string>{});
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -609,8 +619,64 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlNode *FeatureTree) {
...
+    }
+
+    // We ignore the indentation in the beginning of each line since the format
+    // does not force it
+    llvm::StringRef ToStringRef(To);
+
+    if (!ToStringRef.contains(':')) {
+      llvm::errs() << "Failed to read in a constraint since it does not "
+                      "contain a colon.";
+      llvm::errs() << "This violates the format.";
```
```suggestion
      llvm::errs() << "This violates the SXFM format.";
```

lib/Feature/FeatureModelParser.cpp
```diff
@@ -609,8 +619,64 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlNode *FeatureTree) {
...
+
+    // In the following lines, we replace all identifiers by the real feature
+    // name
+    for (auto &Key : Keys) {
+      std::string Value = IdentifierMap[Key];
+      while (CnfFormula.find(Key) != std::string::npos) {
+        auto FoundPos = CnfFormula.find(Key);
+        CnfFormula =
+            CnfFormula.replace(FoundPos, FoundPos + Key.length(), Value);
+      }
```
```suggestion
      auto FoundPos = CnfFormula.find(Key);
      while (FoundPos != std::string::npos) {
        CnfFormula =
            CnfFormula.replace(FoundPos, FoundPos + Key.length(), Value);
        FoundPos = CnfFormula.find(Key);
      }
```

unittests/Feature/FeatureModelSxfm.cpp
```diff
@@ -34,6 +34,16 @@ TEST(SxfmParser, validation) {
             Feature::FeatureKind::FK_BINARY);
 }
 
+/// Here, we check whether a wrongly formatted sxfm file would be parsed
+TEST(FeatureModelParser, detectWornglyFormattedConstraint) {
+  auto FS = llvm::MemoryBuffer::getFileAsStream(
+      getTestResource("test_wrong_constraint_format.sxfm"));
+  EXPECT_TRUE(FS && "Input file could not be read.");
+  auto FM =
+      FeatureModelSxfmParser(FS.get()->getBuffer().str()).buildFeatureModel();
```
vulder:
```suggestion

  auto FM =
      FeatureModelSxfmParser(FS.get()->getBuffer().str()).buildFeatureModel();

```

vulder (changes requested):

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: The identifier is a string such as `(root)`. Therefore, I have to remove the beginning and the ending paranthesis. This is why I need `Identifier.size() - 2`.

lib/Feature/FeatureModelParser.cpp
```diff
@@ -553,6 +553,21 @@ bool FeatureModelSxfmParser::parseFeatureTree(xmlNode *FeatureTree) {
...
+      for (std::sregex_iterator Match = std::sregex_iterator(
+               CurrentLine.begin(), CurrentLine.end(), Regex);
+           Match != std::sregex_iterator(); Match++) {
+        std::string Identifier = (*Match).str();
+        Identifier = Identifier.substr(1, Identifier.size() - 2);
+        IdentifierMap[Identifier] = Name;
+        if (Identifier != Name && Name.find(Identifier) != std::string::npos) {
+          llvm::errs() << "Name must not contain ID:'" << Name << "'\n";
+          return false;
+        }
```
vulder, line 568: Why do you assign this before checking?
If we would do the check before, we could move the name string into the map.

vulder (approved):
However, somehow the coverage tracking does not work. Let's check this after merging. From the website I see only that no base report exists.

lib/Feature/FeatureModelParser.cpp
ChristianKaltenecker: The issue is that I am using the variable `Name` later for creating a feature with this name using the `FeatureModelBuilder`, so `std::move` is no option here. 
But, of course, I could move the check a line earlier.

vulder: kk

PR #69, opened 2021-10-29
Added MemberOffset to XML Parser

Zatho:
Defined XMLConstant MemberOffset in SourceRange
Extended FeatureModelParser to extract MemberOffset as std::string
Extended FeatureSourceRange with std::string MemberOffset

Need feedback particularly on:
- the type conversion xmlChar* to std::string.
- the use of std::string for the MemberOffset, is that what we want?

related se-sic/VaRA#823

### DIFF
--- a/include/vara/Feature/FeatureSourceRange.h
+++ b/include/vara/Feature/FeatureSourceRange.h
@@ -126,7 +126,7 @@ public:
       StrS << "-" << End->toString();
     }
     if (MemberOffset) {
-      StrS << " MemberOffset: " << getMemberOffset();
+      StrS << " MemberOffset: " << &MemberOffset.value();
     }
     return StrS.str();
   }
--- a/include/vara/Feature/FeatureSourceRange.h
+++ b/include/vara/Feature/FeatureSourceRange.h
@@ -74,7 +74,7 @@ public:
                      std::optional<std::string> MemberOffset = std::nullopt,
                      Category CategoryKind = Category::necessary)
       : Path(std::move(Path)), Start(std::move(Start)), End(std::move(End)),
-        MemberOffset(std::move(MemberOffset)) CategoryKind(CategoryKind) {}
+        MemberOffset(std::move(MemberOffset)), CategoryKind(CategoryKind) {}
 
   FeatureSourceRange(fs::path Path, FeatureSourceLocation Start,
                      FeatureSourceLocation End, std::string MemberOffset,
--- a/include/vara/Feature/FeatureSourceRange.h
+++ b/include/vara/Feature/FeatureSourceRange.h
@@ -71,7 +71,7 @@ public:
   FeatureSourceRange(fs::path Path,
                      std::optional<FeatureSourceLocation> Start = std::nullopt,
                      std::optional<FeatureSourceLocation> End = std::nullopt,
-                     std::optional<std::string> MemberOffset = nullptr,
+                     std::optional<std::string> MemberOffset = std::nullopt,
                      Category CategoryKind = Category::necessary)
       : Path(std::move(Path)), Start(std::move(Start)), End(std::move(End)),
         MemberOffset(std::move(MemberOffset)), CategoryKind(CategoryKind) {}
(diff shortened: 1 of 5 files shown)


include/vara/Feature/FeatureSourceRange.h
```diff
@@ -71,15 +71,18 @@ class FeatureSourceRange {
...
+                     std::optional<std::string> MemberOffset = std::nullopt)
       : Path(std::move(Path)), Start(std::move(Start)), End(std::move(End)),
-        CategoryKind(CategoryKind) {}
+        CategoryKind(CategoryKind), MemberOffset(std::move(MemberOffset)) {}
 
   FeatureSourceRange(fs::path Path, FeatureSourceLocation Start,
                      FeatureSourceLocation End,
-                     Category CategoryKind = Category::necessary)
+                     Category CategoryKind = Category::necessary,
+                     std::string MemberOffset = "")
```
vulder: Why do we need two different `MemberOffset` defaults? Having `nullopt` and `""`. Would it not be better to have the same behavior independent on which ctor is called?

unittests/Feature/FeatureSourceRange.cpp
```diff
@@ -90,10 +90,10 @@ TEST(FeatureSourceRange, equality) {
 }
 
 TEST(FeatureSourceRange, clone) {
-  auto FSR = std::make_unique<FeatureSourceRange>(
+  auto FSR = std::make_unique<FeatureSourceRange>(FeatureSourceRange(
```
vulder: Why do we need a direct ctor call here? This would then copy construct the FSR in the unique_ptr with the constructed FSR.

vulder (changes requested):
We should add a few tests that cover the new member offset feature.

include/vara/Feature/FeatureSourceRange.h
```diff
@@ -115,12 +126,16 @@ class FeatureSourceRange {
     if (End) {
       StrS << "-" << End->toString();
     }
+    if (MemberOffset) {
+      StrS << " MemberOffset: " << &MemberOffset.value();
```
vulder: Why is it necessary here to take the address of `MemberOffset`? Should value not just return a string?

vulder (commented):
Overall, the current state looks fine to me.


@s9latimm Should be add a specific class to represent that MemberOffset?
Could be helpful, as it makes accessing the data easier, e.g.:
```
class FeatureMemberOffset {
  llvm::StringRef className() const;
  llvm::StringRef memberName() const;
};
```

+ Do you think we want/need nesting? I guess, the most common case will be without nesting, so just implementing `Class::Member` should be enough.

s9latimm:
LGTM

> Overall, the current state looks fine to me.
> 
> @s9latimm Should be add a specific class to represent that MemberOffset? Could be helpful, as it makes accessing the data easier, e.g.:
> 
> ```
> class FeatureMemberOffset {
>   llvm::StringRef className() const;
>   llvm::StringRef memberName() const;
> };
> ```
> 
>     * Do you think we want/need nesting? I guess, the most common case will be without nesting, so just implementing `Class::Member` should be enough.

Having this information allow us to easily dump the offset and probably makes  processing / compiling with a FM easier, but I am not sure yet, if we really need this though.

On the other hand I think we are currently fine without fully supporting nested members.

s9latimm (approved):

include/vara/Feature/FeatureSourceRange.h
```diff
@@ -68,18 +68,52 @@ class FeatureSourceRange {
...
-                     Category CategoryKind = Category::necessary)
+  class FeatureMemberOffset {
+  public:
+    FeatureMemberOffset(std::string MemberOffset)
+        : MemberOffset(std::move(MemberOffset)) {}
+    FeatureMemberOffset(const FeatureMemberOffset &L) = default;
+    FeatureMemberOffset &operator=(const FeatureMemberOffset &) = default;
+    FeatureMemberOffset(FeatureMemberOffset &&) = default;
+    FeatureMemberOffset &operator=(FeatureMemberOffset &&) = default;
+    virtual ~FeatureMemberOffset() = default;
```
vulder:
```suggestion
```
We only need to provide the special member functions, in case we change one.

vulder (approved):

PR #70, opened 2021-11-02
Fixes vara namespace for own Result/Error utilities

vulder:


unittests/Utils/Result.cpp
```diff
@@ -4,6 +4,11 @@
 
 namespace test {
 
+using vara::Ok;
+using vara::Error;
+using vara::Ok;
```
vulder, line 8:
```suggestion
```

unittests/Utils/Result.cpp
```diff
@@ -4,6 +4,10 @@
 
 namespace test {
 
+using vara::Ok;
```
vulder:
```suggestion
using vara::Error;
using vara::Ok;
using vara::Result;
```

boehmseb (approved):

PR #71, opened 2021-11-30
Implements configuration class

ChristianKaltenecker:


include/vara/Configuration/ConfigurationOption.h
```diff
@@ -0,0 +1,49 @@
+#ifndef VARA_CONFIGURATION_CONFIGURATIONOPTION_H
```
vulder: I don't think we need to put every class in a separate header. The config classes are all tied together and, hence, are included together, so we can put them in one header file.

include/vara/Configuration/Configuration.h
```diff
@@ -0,0 +1,50 @@
+#ifndef VARA_CONFIGURATION_CONFIGURATION_H
+#define VARA_CONFIGURATION_CONFIGURATION_H
+
+#include "vara/Configuration/ConfigurationOption.h"
+
+namespace vara::feature {
+
+/// \brief This class represents a configuration.
+/// The configuration is represented by using a json format of the form
```
vulder:
```suggestion
/// \brief This class represents a configuration.
///
/// The configuration is represented by using a json format of the form
```

include/vara/Configuration/Configuration.h
```diff
@@ -0,0 +1,50 @@
...
+/// \c `{name: value, name2: value2}`.
+class Configuration {
+public:
+  explicit Configuration() = default;
```
vulder:
```suggestion
  Configuration() = default;
```
`explicit` is mostly important for ctors that take 1 parameter, but I don't think we need it in this case here.

include/vara/Configuration/Configuration.h
```diff
@@ -0,0 +1,50 @@
...
+  virtual ~Configuration() = default;
+
+  /// This method creates a configuration from the provided json string.
+  [[nodiscard]] static std::shared_ptr<Configuration>
+  createConfigurationFromString(const std::string &ConfigurationString);
+
+  /// This method adds a configuration option to the current configuration.
+  void addConfigurationOption(std::shared_ptr<ConfigurationOption> Option);
```
vulder: Why do we need a `shared_ptr` here? The option will only be inside of the Configuration so using a `unique_ptr` is the better choice.

include/vara/Configuration/Configuration.h
```diff
@@ -0,0 +1,50 @@
...
+  getConfigurationOptions();
+
+  /// This method dumps the current configuration to a json string.
+  /// \returns the current configuration as a json-formatted string
+  [[nodiscard]] std::string dumpToString();
+
+private:
+  /// This represents a mapping from the name of the option to the option
+  /// object.
+  std::map<std::string, std::shared_ptr<ConfigurationOption>> OptionMapping;
```
vulder: Use `unique_ptr`

include/vara/Configuration/Configuration.h
```diff
@@ -0,0 +1,50 @@
...
+
+  /// This method returns the value of the configuration option.
+  /// \returns the value of the configuration option as a string
+  [[nodiscard]] std::string
+  getConfigurationOptionValue(const std::string &Name);
+
+  /// This method returns all configuration options.
+  /// \returns all configuration options in a vector
+  [[nodiscard]] std::vector<std::shared_ptr<ConfigurationOption>>
+  getConfigurationOptions();
```
vulder: Why do we need to return them all there? An iterator that allows iteration over the options is enough for all users to access them.

include/vara/Configuration/ConfigurationOption.h
```diff
@@ -0,0 +1,49 @@
...
+#define VARA_CONFIGURATION_CONFIGURATIONOPTION_H
+
+#include <llvm/Support/FormatVariadic.h>
+
+#include <map>
+#include <string>
+#include <vector>
+
+using std::string;
```
vulder: Pls try not o use using declarations in header files, as they are included in user code and force users to deal with naming problems/name clashes.

include/vara/Configuration/ConfigurationOption.h
```diff
@@ -0,0 +1,49 @@
...
+      : Name(std::move(Name)), Value(std::move(Value)){};
+
+  /// This method returns the name of the configuration option.
+  /// \returns the name of the configuration option as a string
+  [[nodiscard]] std::string getName() { return this->Name; }
+
+  /// This method returns the value of the configuration option.
+  /// \returns the value of the configuration option as a string
+  [[nodiscard]] std::string getValue() { return this->Value; }
+
```
vulder: We can just return `std::string_view` here so we don't need to copy the strings all the time.

include/vara/Configuration/ConfigurationOption.h
```diff
@@ -0,0 +1,49 @@
...
+  [[nodiscard]] std::string toString() {
+    return llvm::formatv("{0}: {1}", this->Name, this->Value);
+  }
+
+private:
+  /// The name of the configuration option
+  const std::string Name;
+
+  /// The value of the configuration option
+  const std::string Value;
```
vulder: Should we handle everything as string? In the configs/json level we would also have int/bool as types.

lib/Configuration/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+
+set(LLVM_LINK_COMPONENTS
+  Support
+  Demangle
+  Core
+  )
+
+add_vara_library(VaRAConfiguration
+  ${CONFIGURATION_LIB_SRC}
+  )
```
vulder:
```suggestion
)

set(LLVM_LINK_COMPONENTS
  Support
  Demangle
  Core
)

add_vara_library(VaRAConfiguration
  ${CONFIGURATION_LIB_SRC}
)
```

lib/Configuration/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+
+if (HAS_STD_EXPERIMENTAL_FILESYSTEM OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  target_link_libraries(VaRAConfiguration
+    LINK_PUBLIC
+    stdc++fs
+    )
```
vulder: For what is this needed?

lib/Configuration/Configuration.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include "llvm/Support/JSON.h"
+
+#include <sstream>
+
+namespace vara::feature {
+
+std::shared_ptr<Configuration> Configuration::createConfigurationFromString(
+    const std::string &ConfigurationString) {
+  std::shared_ptr<Configuration> Conf = std::make_unique<Configuration>();
```
vulder: The right side tells what should be :rofl: 
`shared_ptr` -> `unique_ptr`

lib/Configuration/Configuration.cpp
```diff
@@ -0,0 +1,88 @@
...
+  // Read in the string using the json library
+  llvm::Expected<llvm::json::Value> ParsedConfiguration =
+      llvm::json::parse(llvm::StringRef(ConfigurationString));
```
vulder: If we need a `StringRef` here, we could just adapt the functions parameter to `StringRef` instead of const std::string &

lib/Configuration/Configuration.cpp
```diff
@@ -0,0 +1,88 @@
...
+  }
+  llvm::json::Value Value = ParsedConfiguration.get();
+  llvm::json::Value::Kind Kind = Value.kind();
+  if (Kind != llvm::json::Value::Kind::Object) {
+    llvm::errs() << "The provided json has to be a map from the name to the "
+                    "value (as string).\n";
+    return nullptr;
+  }
+  llvm::json::Object *Obj = Value.getAsObject();
+  for (auto Iterator = Obj->begin(); Iterator != Obj->end(); Iterator++) {
```
vulder: Do you need an iterator loop here? A for each loop would be simpler.

lib/Configuration/Configuration.cpp
```diff
@@ -0,0 +1,88 @@
...
+    std::string First = Iterator->getFirst().str();
+    if (Iterator->getSecond().kind() != llvm::json::Value::Kind::String) {
+      llvm::errs() << "The values of the provided json string have to be "
+                      "simple strings.\n";
+      return nullptr;
+    }
+    std::string Second = Iterator->getSecond().getAsString()->str();
+    std::shared_ptr<ConfigurationOption> Option =
+        std::make_unique<ConfigurationOption>(First, Second);
```
vulder: Use `unique_ptr`

*Just as a side node, you always use `make_unique` to create `shared_ptrs`, `make_unique` is for creating `unique_ptrs`, `make_shared` is for creating `shared_ptrs`

lib/Configuration/Configuration.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+void Configuration::addConfigurationOption(
+    std::shared_ptr<ConfigurationOption> Option) {
+  this->OptionMapping[Option->getName()] = std::move(Option);
+}
+
+void Configuration::setConfigurationOption(const std::string &Name,
+                                           const std::string &Value) {
+  std::shared_ptr<ConfigurationOption> Option =
```
vulder: same

lib/Configuration/Configuration.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+std::string
+Configuration::getConfigurationOptionValue(const std::string &Name) {
+  auto Iterator = this->OptionMapping.find(Name);
+  if (Iterator == this->OptionMapping.end()) {
+    llvm::errs() << "The configuration option " << Name << " does not exist.\n";
+    return "";
+  }
+  return Iterator->second->getValue();
```
vulder:
```suggestion
  auto Search = this->OptionMapping.find(Name);
  if (Search == this->OptionMapping.end()) {
    llvm::errs() << "The configuration option " << Name << " does not exist.\n";
    return "";
  }
  return Search->second->getValue();
```

lib/Configuration/Configuration.cpp
```diff
@@ -0,0 +1,88 @@
...
+    Options.insert(Options.begin(), Iterator.second);
+  }
+  return Options;
+}
+
+std::string Configuration::dumpToString() {
+  llvm::json::Object Obj{};
+  for (auto &Iterator : this->OptionMapping) {
+    Obj[Iterator.first] = Iterator.second->getValue();
+  }
```
vulder:
```suggestion
  for (auto &OptionMapping : this->OptionMappings) {
    Obj[Iterator.first] = OptionMapping.second->getValue();
  }
```
Should the original `OptionMapping` not be named `OptionMappings` as there are multiple and so we need plural?

unittests/CMakeLists.txt
```diff
@@ -1,14 +1,17 @@
 add_custom_target(VaRAFeatureUnitTests)
+add_custom_target(VaRAConfigurationUnitTests)
```
vulder: I think you miss understood my intention with the `VaRAFeatureUnitTests`
The idea was that this represents ALL vara-feature unittests, meaning configuration tests would be in a subfolder/substructure of this unittest package.

Maybe we should restructure the tests a bit, it's currently not 100% clear as up to know everything was called feature and "dumped" into the same folder. I'll try to have a look at this, when I find time.

unittests/Configuration/Configuration.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "gtest/gtest.h"
+
+namespace vara::feature {
+TEST(Configuration, basicAccessors) {
+  std::string ConfigurationString = R"({"baz":"1","foo":"true"})";
+  auto Config =
+      Configuration::createConfigurationFromString(ConfigurationString);
+  EXPECT_TRUE(Config);
+  std::string Dump = Config->dumpToString();
```
vulder: I think we should split this test into two, as it tests parsing and dumping.
Test1 should check if we can correctly parse config files.
Where another one, should test if we can correctly dump existing configs.

Maybe also have 1-2 different tests for different scenarios (should there be a bit more complicated ones).

vulder (changes requested):

include/vara/Configuration/Configuration.h
ChristianKaltenecker: At first, I have indeed used a unique_ptr. 
This, however, does not work because of the `getConfigurationOptions()`-method which returns the configuration options.
Should I copy the configuration options in the method `getConfigurationOptions()` and return them?

unittests/CMakeLists.txt
ChristianKaltenecker: So, should I move the tests inside the vara-feature folder?

include/vara/Configuration/Configuration.h
ChristianKaltenecker: Because of the `setOptions(self)` method in https://github.com/se-sic/VaRA-Tool-Suite/blob/vara-dev/varats-core/varats/base/configuration.py
There we return a list of all configuration options.
A user relying on the future python binding of the `Configuration` class should also have the possibility to get all configuration options, right?

Are there maybe better ways to do this?

include/vara/Configuration/ConfigurationOption.h
ChristianKaltenecker: You are right.
The `ConfigurationOption` class is merely a container of the name and value without any further checks.
However, I intended to add a verify method that uses further knowledge regarding the configuration option.

lib/Configuration/CMakeLists.txt
ChristianKaltenecker: Will be removed. This was a mistake.

vulder (approved):

PR #72, opened 2021-12-06
Refactors test detection strategy to support multiple test units

vulder:


ChristianKaltenecker (approved):
Looks good to me. Thanks for fixing this!

PR #74, opened 2021-12-10
Parse Large Numbers and Scientific Notation

s9latimm:
Allow numbers to be written in scientific notation in XML / constraints and handle out-of-range integers.

related se-sic/VaRA/issues/813

closes se-sic/VaRA/issues/814
closes se-sic/VaRA/issues/815


include/vara/Feature/Constraint.h
```diff
@@ -570,10 +570,10 @@ class PrimaryConstraint : public Constraint {
 
 class PrimaryIntegerConstraint : public PrimaryConstraint {
 public:
-  PrimaryIntegerConstraint(int Value)
+  PrimaryIntegerConstraint(long Value)
```
```suggestion
  PrimaryIntegerConstraint(int64_t Value)
```

I would use `int64_t` instead
a) it makes the size clear
b) it also works for 32bit latforms

include/vara/Feature/ConstraintParser.h
```diff
@@ -230,7 +230,10 @@ class ConstraintLexer {
   }
 
   static ResultTy munchNumber(const llvm::StringRef &Str) {
-    auto Munch = Str.take_while([](auto C) { return llvm::isDigit(C); });
+    auto Munch = Str.take_while([](auto C) {
+      return llvm::isDigit(C) || C == 'e' || C == '+' || C == '-';
```
vulder: did you run `toLower` before? If not, 'E' could also happen

`12E3`

include/vara/Feature/ConstraintParser.h
```diff
@@ -497,6 +499,30 @@ class ConstraintParser {
     }
   }
 
+  static std::unique_ptr<Constraint> parseUnsigned(llvm::StringRef Str) {
+    if (Str.contains_lower('e')) {
```
vulder: Did they rename the method? In the StringRef API I only find `contains_insensitive` and not `contains_lower`

include/vara/Feature/ConstraintParser.h
```diff
@@ -497,6 +499,30 @@ class ConstraintParser {
...
+    try {
+      return std::make_unique<PrimaryIntegerConstraint>(std::stol(Str.str()));
+    } catch (std::invalid_argument &_) {
+      llvm::errs() << "Failed to parse number '" << Str << "'\n";
+      return nullptr;
+    } catch (std::out_of_range &_) {
+      llvm::errs() << "Number too large.\n";
+      return std::make_unique<PrimaryIntegerConstraint>(
+          std::numeric_limits<long>::max());
+    }
```
vulder: Do we have a way to implement this without exception handling? LLVM builds without exceptions and vara-feature is embedded into llvm/vara

lib/Feature/FeatureModelParser.cpp
```diff
@@ -123,6 +123,32 @@ FeatureModelXmlParser::parseConfigurationOption(xmlNode *Node,
...
+    llvm::errs() << "Failed to parse number '" << Str << "'\n";
+    return 0;
+  } catch (std::out_of_range &_) {
+    llvm::errs() << "Number out of range.\n";
+    if (Str.startswith("-")) {
+      return std::numeric_limits<long>::min();
+    }
+    return std::numeric_limits<long>::max();
+  }
+}
```
vulder: This seems to be a duplicate of ConstrantParser `parseUnsigned`, we should refactor this into a single impl

vulder (changes requested):

include/vara/Feature/ConstraintParser.h
s9latimm: This method was only renamed six months ago (https://github.com/llvm/llvm-project/commit/3eed57e7ef7da5eda765ccc19fd26fb8dfcd8d41), therefore i use the old name with our LLVM version.
This will probably break at some point, as we update the framework.

vulder: :+1: I thought something like this :)

vulder (approved):

include/vara/Feature/ConstraintParser.h
```diff
@@ -257,6 +260,31 @@ class ConstraintLexer {
...
+  int64_t Integer;
+  if (!Str.getAsInteger(10, Integer)) {
+    return Integer;
+  }
+
+  if (Str.contains_lower('e')) {
+    // If we encounter scientific notation we try to parse the number as double.
+    double Double = 0;
+    if (!Str.getAsDouble(Double)) {
+      return round(Double);
```
vulder: We need to add a check here, to war should a user pass an value that's to large.

vulder (approved):

PR #79, opened 2022-01-13
Updates FeatureModelOffset to enforce the desired format

vulder:


s9latimm (approved):
LGTM

Using the name of the desired member would work fine with my draft to add metadata in Clang.

Do we store the location of the enclosing struct or the field declaration itself?

PR #80, opened 2022-01-14
Add Parent and Child Within One Commit

s9latimm:
Closes https://github.com/se-sic/VaRA/issues/799


vulder:
Currently, declined as the feature would not improve readability.

PR #84, opened 2022-02-01
Detect whether a linker flag is needed for std::filesystem.

boehmseb:


cmake/vara_feature_macros.cmake
```diff
@@ -82,3 +82,48 @@ int main() { return 0; }
...
+  # Check if we need to add -lstdc++fs or -lc++fs or nothing
+  set(OLD_CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES})
+  set(OLD_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+  set(CMAKE_REQUIRED_FLAGS "-std=c++17 ${CMAKE_REQUIRED_FLAGS}")
+
+  CHECK_CXX_SOURCE_COMPILES("
+#include <filesystem>
+int main(int argc, char ** argv) {
+std::filesystem::path p(argv[0]);
+return p.string().length(); }
```
vulder, line 96: Can you move the duplicated code into a local variable, so we share it between all three compile checks?

vulder (changes requested):
Otherwise, this looks fine :)

cmake/vara_feature_macros.cmake
boehmseb: I tried that with `set(TEST_PROG "...")` but then `CHECK_CXX_SOURCE_COMPILES(${TEST_PROG} TEST_RESULT)` would always give an error.

vulder: But would it not be `CHECK_CXX_SOURCE_COMPILES("${TEST_PROG}" TEST_RESULT)`

I think this should work, but maybe I'm wrong an CHECK_CXX_SOURCE is special somehow?

boehmseb: Hmm, the extra `""` could be the solution.

vulder (approved):

PR #90, opened 2022-06-03
Solver

ChristianKaltenecker:
Provides means to retrieve all configurations of a feature model.

- [x] Read in features from the feature model
- [x] Support parent-child relationship
- [x] Parse constraints
- [x] Provide an iterator over the configurations
- [x] Provide a solver factory
- [x] Incorporate feedback from Florian
- [x] Change the way how z3 is currently included into the package (since the official z3 package produces Segfaults in Debian 11)


include/vara/Solver/ConfigurationFactory.h
```diff
@@ -0,0 +1,133 @@
...
+#include <vector>
+
+namespace vara::solver {
+
+/// This class represents a configuration iterator that lazily traverses
+/// over all configurations.
+class ConfigurationIterator {
+public:
+  class Iterator {
+  public:
```
vulder, line 19:
```suggestion
  public:
      using iterator_category = ?;
      using value_type = ?;
      using difference_type = ?;
      using pointer = ?;
      using reference = ?;
```
Should define iterator traits to be easy usable with standard algorithms

include/vara/Solver/ConfigurationFactory.h
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    bool operator==(Iterator Other) const { return S == Other.S; }
+    bool operator!=(Iterator Other) const { return !(Other == *this); }
+
+  private:
+    Solver *S;
+  };
+
+  explicit ConfigurationIterator(std::unique_ptr<Solver> Solver)
```
vulder: What is the idea behind the nested iterator, producing two different types?

include/vara/Solver/ConfigurationFactory.h
```diff
@@ -0,0 +1,133 @@
...
+};
+
+/// This class is used to retrieve information about the feature model
+/// (i.e., how many configurations are there? is it valid?)
+class ConfigurationFactory {
+public:
+  /// This method returns an iterator. This iterator can be used to lazily
+  /// traverse over configurations retrieved by the solver.
+  /// \param Model the given model containing the features and constraints
+  /// \param Type the type of solver to use
```
vulder, line 71:
```suggestion

  /// \param Model the given model containing the features and constraints
  /// \param Type the type of solver to use
  
```

include/vara/Solver/ConfigurationFactory.h
```diff
@@ -0,0 +1,133 @@
...
+  /// \return A unique pointer to the configuration iterator
```
vulder:
```suggestion
  /// \returns A unique pointer to the configuration iterator
```

include/vara/Solver/ConfigurationFactory.h
```diff
@@ -0,0 +1,133 @@
...
+        new ConfigurationIterator(std::move(S)));
+    return Iterator;
+  }
+
+  /// This method returns all configurations of the given feature model. Note
+  /// that retrieving all configurations might take some time or even be
+  /// infeasible, especially for larger feature models.
+  /// \param Model the given model containing the features and constraints
+  /// \param Type the type of solver to use
+  /// \return a unique pointer to the vector containing all configurations
```
vulder:
```suggestion

  /// \param Model the given model containing the features and constraints
  /// \param Type the type of solver to use
  
  /// \returns a unique pointer to the vector containing all configurations
```
same for the rest

include/vara/Solver/ConfigurationFactory.h
```diff
@@ -0,0 +1,133 @@
...
+  /// traverse over configurations retrieved by the solver.
+  /// \param Model the given model containing the features and constraints
+  /// \param Type the type of solver to use
+  /// \return A unique pointer to the configuration iterator
+  static std::unique_ptr<ConfigurationIterator>
+  getConfigIterator(feature::FeatureModel &Model,
+                    const vara::solver::SolverType Type = SolverType::Z3) {
+    auto S = SolverFactory::initializeSolver(Model, Type);
+    std::unique_ptr<ConfigurationIterator> Iterator(
+        new ConfigurationIterator(std::move(S)));
```
vulder:
```suggestion
    auto Iterator = std::make_unique<ConfigurationIterator>(std::move(S));
```

include/vara/Solver/ConfigurationFactory.h
```diff
@@ -0,0 +1,133 @@
...
+  /// \param Type the type of the solver
+  /// \return a unique pointer to a vector containing the specified amount of
+  /// configurations.
+  static Result<SolverErrorCode,
+                std::unique_ptr<
+                    std::vector<std::unique_ptr<vara::feature::Configuration>>>>
+  getNConfigs(feature::FeatureModel &Model, uint N,
+              const vara::solver::SolverType Type = SolverType::Z3) {
+    auto Iterator = getConfigIterator(Model, Type);
+    std::unique_ptr<std::vector<std::unique_ptr<feature::Configuration>>> V =
```
vulder: Do we need to store the vector it self also on the heap? Would be easier to just return it.
-> returning the vector by value would construct it in place for the caller, no copy needed (RVO)
+ If one want's to move it around it can just be moved, as moving a vector is cheap.

include/vara/Solver/Error.h
```diff
@@ -0,0 +1,65 @@
...
+} // namespace solver
+
+template <>
+class Error<vara::solver::SolverErrorCode> {
+public:
+  Error(vara::solver::SolverErrorCode E) : E(E) {}
+
+  vara::solver::SolverErrorCode operator*() { return E; }
+
+  vara::solver::SolverErrorCode extractError() { return E; }
```
vulder, line 28: This part looks quite generic. @s9latimm is this necessary? Maybe we could find a design where a specialization does not have to reimplement the trivial members :thinking:

include/vara/Solver/Error.h
```diff
@@ -0,0 +1,65 @@
...
+
+  operator bool() const { return false; }
```
vulder, line 30: That looks wrong, doesn't it?

include/vara/Solver/Solver.h
```diff
@@ -0,0 +1,252 @@
...
+  virtual Result<SolverErrorCode, std::unique_ptr<std::vector<std::unique_ptr<
+                                      vara::feature::Configuration>>>>
+  getAllValidConfigurations() = 0;
+};
+
+//===----------------------------------------------------------------------===//
+//                               Z3Solver Class
+//===----------------------------------------------------------------------===//
+
+/// \brief The Z3 solver implementation for handling constraints
```
vulder: Maybe we should put this into a different file.

include/vara/Solver/Solver.h
```diff
@@ -0,0 +1,252 @@
...
+  z3::context Context;
+
+  /// The instance of the Z3 solver needed for caching the constraints and
+  /// variables.
+  std::unique_ptr<z3::solver> Solver;
+};
+
+/// \brief This class is a visitor to convert the constraints from the
+/// feature model into constraints for Z3.
+class SolverConstraintVisitor : public vara::feature::ConstraintVisitor {
```
vulder:
```suggestion
class Z3SolverConstraintVisitor : public vara::feature::ConstraintVisitor {
```
If it's Z3 specific, maybe the name should reflect that.

lib/Solver/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+)
+
+add_vara_library(VaRASolver
+  ${SOLVER_LIB_SRC}
+)
+
+target_link_libraries(VaRASolver
+        LINK_PUBLIC
+        VaRAConfiguration
+        )
```
vulder:
```suggestion
)
```

lib/Solver/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+set(LLVM_LINK_COMPONENTS
+)
+
+add_vara_library(VaRASolver
+  ${SOLVER_LIB_SRC}
+)
+
+target_link_libraries(VaRASolver
+        LINK_PUBLIC
+        VaRAConfiguration
```
vulder:
```suggestion
  LINK_PUBLIC
  VaRAConfiguration
```

lib/Solver/SolverFactory.cpp
```diff
@@ -0,0 +1,35 @@
...
+  for (const auto &C : Model.constraints()) {
+    S->addConstraint(*C);
+  }
+
+  // Iterate over all relationships
+  for (const auto &R : Model.relationships()) {
+    S->addRelationship(*R);
+  }
+  return S;
+}
```
vulder: If I read this correctly, we could split this function into 2. 1x where the solver ist created and a 1x where the FeatureModel is applied to the created solver, as the second one only uses the highlevel solver interface and, therefore could be reused for other solvers.

unittests/Solver/BasicSolverTests.cpp
```diff
@@ -0,0 +1,141 @@
...
+  z3::expr Expression = A && B;
+
+  z3::solver S(C);
+
+  S.add(Expression);
+  EXPECT_EQ(S.check(), z3::check_result::sat);
+
+  z3::expr InvalidatingExpression = !A;
+  S.add(InvalidatingExpression);
+  EXPECT_EQ(S.check(), z3::check_result::unsat);
```
vulder, line 23: I'm not 100% sure if this test checks what I think it does.
`A && B !A` is not a correct formula (which matches the name InvalidExpression)
However, we still expect z3 to return UNSAT? Should this not be an error?

unittests/Solver/BasicSolverTests.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+  z3::model M = S.get_model();
+
+  // Retrieve the model by using the declaration function of each expression
+  EXPECT_TRUE(M.get_const_interp(A.decl()));
+  EXPECT_TRUE(M.get_const_interp(B.decl()));
+  EXPECT_FALSE(M.get_const_interp(C.decl()));
+
+  // Add the current configuration as a constraint
+  if (M.get_const_interp(D.decl())) {
```
vulder: Why is there a if in the test? How does D.decl change as the rest of the test does not influence it direclty.

unittests/Solver/ConfigurationFactory.cpp
```diff
@@ -0,0 +1,84 @@
...
+  B.makeFeature<feature::BinaryFeature>("b", true);
+  B.addEdge("alt", "a");
+  B.addEdge("alt", "b");
+  std::unique_ptr<feature::Constraint> C =
+      std::make_unique<feature::ImpliesConstraint>(
+          std::make_unique<feature::PrimaryFeatureConstraint>(
+              std::make_unique<feature::BinaryFeature>("a")),
+          std::make_unique<feature::NotConstraint>(
+              std::make_unique<feature::PrimaryFeatureConstraint>(
+                  std::make_unique<feature::BinaryFeature>("b"))));
```
vulder: That ones quick complicated to write, @s9latimm maybe we should at some point simplify this synatx

unittests/Solver/RealWorldCaseStudyTests.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "llvm/ADT/Twine.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "gtest/gtest.h"
+
+namespace vara::solver {
+
+std::unique_ptr<vara::feature::FeatureModel>
+buildFeatureModel(llvm::StringRef Path) {
+  auto FS = llvm::MemoryBuffer::getFileAsStream(Path);
+  // auto FS = llvm::MemoryBuffer::getFileAsStream(getTestResource(Path));
```
vulder: Does getTestResource not work?

unittests/Solver/RealWorldCaseStudyTests.cpp
```diff
@@ -0,0 +1,67 @@
...
+  // Perform a test on Dune with numeric features with step functions
+  auto FMNumStep = buildFeatureModel(
+      "../../../unittests/resources/xml/test_dune_num_explicit.xml");
+  auto ConfigResultNumStep = ConfigurationFactory::getAllConfigs(*FMNumStep);
+  EXPECT_TRUE(ConfigResultNumStep);
+  EXPECT_EQ(ConfigResultNumStep.extractValue()->size(), 2304);
+}
+
+TEST(SolverAPI, TestHipaccConfigurations) {
+  GTEST_SKIP();
```
vulder: So there was no direct way to label/tag the test?

unittests/resources/xml/vm.dtd
```diff
@@ -5,7 +5,7 @@
 <!ELEMENT booleanConstraints (constraint*)>
 <!ELEMENT nonBooleanConstraints (constraint*)>
 <!ELEMENT mixedConstraints (constraint*)>
-<!ELEMENT configurationOption (name, outputString?, (prefix | preFix)?, (postfix | postFix)?, parent?, children?, impliedOptions?, excludedOptions?, defaultValue?, optional?, ((minValue, maxValue) | values)?, stepFunction?, locations?)>
+<!ELEMENT configurationOption (name, outputString?, (prefix | preFix)?, (postfix | postFix)?, parent?, children?, impliedOptions?, excludedOptions?, optional?, ((minValue, maxValue) | values)?, stepFunction?, defaultValue?, locations?)>
```
vulder, line 8: Why did this change here?

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+  return Ok();
+}
+
+Result<SolverErrorCode>
+Z3Solver::addFeature(const string &FeatureName,
+                     const std::vector<int64_t> &Values) {
+  if (OptionToVariableMapping.find(llvm::StringRef(FeatureName)) !=
+      OptionToVariableMapping.end()) {
+    return ALREADY_PRESENT;
+  }
```
vulder: x

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+  Solver->push();
+  return Ok();
+}
+
+Result<SolverErrorCode> Z3Solver::addFeature(const string &FeatureName) {
+  if (OptionToVariableMapping.find(FeatureName) !=
+      OptionToVariableMapping.end()) {
+    return ALREADY_PRESENT;
+  }
```
vulder: x

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+  vara::feature::Feature *Parent = FeatureToAdd.getParentFeature();
+  if (Parent != nullptr && OptionToVariableMapping.find(Parent->getName()) ==
+                               OptionToVariableMapping.end()) {
+    return PARENT_NOT_PRESENT;
+  }
+
+  if (OptionToVariableMapping.find(FeatureToAdd.getName()) !=
+      OptionToVariableMapping.end()) {
+    return ALREADY_PRESENT;
+  }
```
vulder: This and the two copies below (X) could be a helper function.

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+      auto Step = Range.first;
+      std::vector<int64_t> Vals;
+      while (Step < Range.second) {
+        Vals.insert(Vals.begin(), Step);
+        Step = StepFunction->next(Step);
+      }
+      if (auto R = addFeature(F->getName().str(), Vals); !R) {
+        return R;
+      }
+    }
```
vulder: You could you std::visit here (maybe together wit Overloaded and a lambda or a helper function).
Visit is a bit more straight forward here and helps us to identify problems should the type space of the variant change.

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+    break;
+  }
+  case feature::Feature::FeatureKind::FK_BINARY: {
+    addFeature(FeatureToAdd.getName().str());
+    // Add all constraints (i.e., implications, exclusions, parent feature)
+    if (!llvm::dyn_cast<vara::feature::BinaryFeature>(&FeatureToAdd)) {
```
vulder:
```suggestion
    if (!llvm::isa<vara::feature::BinaryFeature>(&FeatureToAdd)) {
```
I would also move this above the `addFeature` call an check as early as possible

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+        IsInAlternativeGroup);
+    if (!R) {
+      return R;
+    }
+    break;
+  }
+  case feature::Feature::FeatureKind::FK_ROOT:
+    addFeature(FeatureToAdd.getName().str());
+    // Add root as a constraint; root is mandatory
+    Solver->add(*OptionToVariableMapping[FeatureToAdd.getName()]);
```
vulder:
```suggestion
    assert(OptionToVariableMapping.find(FeatureToAdd.getName()) != OptionToVariableMapping.end() && "No OptionToVariableMapping was defined for the given feature.");
    Solver->add(*OptionToVariableMapping[FeatureToAdd.getName()]);
```

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+}
+
+Result<SolverErrorCode>
+Z3Solver::removeFeature(feature::Feature &FeatureToRemove) {
+  return NOT_SUPPORTED;
+}
+
+Result<SolverErrorCode>
+Z3Solver::addRelationship(const feature::Relationship &R) {
+  const auto *Parent = (const feature::Feature *)R.getParent();
```
vulder: Please don't use c-style casts.

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+Z3Solver::addConstraint(feature::Constraint &ConstraintToAdd) {
+  SolverConstraintVisitor SCV(this);
+  bool Succ = SCV.addConstraint(&ConstraintToAdd);
+  if (!Succ) {
+    return SolverErrorCode::NOT_SUPPORTED;
+  }
+  return Ok();
+}
+
+Result<SolverErrorCode, std::unique_ptr<bool>>
```
vulder: Why is the bool here store in a unique_ptr and not just copied?

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+Result<SolverErrorCode, std::unique_ptr<vara::feature::Configuration>>
+Z3Solver::getNextConfiguration() {
+  // Add previous configuration as a constraint
+  excludeCurrentConfiguration();
+
+  // Retrieve the next configuration
+  return getCurrentConfiguration();
+}
+
+Result<SolverErrorCode, std::unique_ptr<uint64_t>>
```
vulder: same here

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+}
+
+Result<SolverErrorCode> Z3Solver::excludeCurrentConfiguration() {
+  if (Solver->check() == z3::unsat) {
+    return UNSAT;
+  }
+  z3::model M = Solver->get_model();
+  z3::expr Expr = Context.bool_val(false);
+  for (auto Option : OptionToVariableMapping.keys()) {
+    z3::expr OptionExpr = *OptionToVariableMapping[Option];
```
vulder: If you need both, key and value normally iterate over the map using the iterator then you have direct access to both and don't have to lookup the value later on.

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+}
+
+Result<SolverErrorCode, std::unique_ptr<vara::feature::Configuration>>
+Z3Solver::getCurrentConfiguration() {
+  if (Solver->check() == z3::unsat) {
+    return UNSAT;
+  }
+  z3::model M = Solver->get_model();
+  std::unique_ptr<vara::feature::Configuration> Config(
+      new vara::feature::Configuration());
```
vulder:
```suggestion
  auto Config = std::make_unique<vara::feature::Configuration>();
```

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+  for (auto Option : OptionToVariableMapping.keys()) {
+    z3::expr OptionExpr = *OptionToVariableMapping[Option];
```
vulder: same here

lib/Solver/Solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+    break;
+  case feature::Constraint::ConstraintKind::CK_SUBTRACTION:
+    Z3ConstraintExpression = Left - Z3ConstraintExpression;
+    break;
+  case feature::Constraint::ConstraintKind::CK_XOR:
+    Z3ConstraintExpression = z3::implies(Left, !Z3ConstraintExpression) &&
+                             z3::implies(Z3ConstraintExpression, !Left);
+    break;
+  default:
+    return false;
```
vulder:
```suggestion
    return false;
```
Should this case ever happen? Maybe we should make this clear and not handle the case silently?

vulder (changes requested):

unittests/Solver/RealWorldCaseStudyTests.cpp
ChristianKaltenecker: It does work; but not in my IDE^^ This is something I will fix in the next revision.

unittests/Solver/RealWorldCaseStudyTests.cpp
ChristianKaltenecker: There is one by applying a filter. This is also something I will change in the next revision.

unittests/resources/xml/vm.dtd
ChristianKaltenecker: Because the original XML format from SPL Conqueror has the defaultValue tag before the optional tag for boolean features and after the stepFunction for numeric features.
From a perspective of a numeric feature it might be better to first see which values a numeric feature can have and afterwards see the default value.
The other alternative would be to undo my change and modify the feature models so that the defaultValue tag is always before optional.

unittests/Solver/BasicSolverTests.cpp
ChristianKaltenecker: If I remember correctly, it could be that the solver (depending on the seed) could find a configuration in different order. But I will try to control this and remove the if in this test.

unittests/Solver/BasicSolverTests.cpp
ChristianKaltenecker: This is just a basic test of Z3 functionality and not of the Z3 integration into VaRA, where we usually return an error if the feature model is unsatisfiable.

Correct; this formula is a contradiction in itself and therefore, Z3 should not find any solution to that (i.e., the solver returns `unsat`.
Typically, a solver has only two outputs: `sat` or `unsat`.

lib/Solver/SolverFactory.cpp
ChristianKaltenecker: This should be possible, yes. I will do this in my next revision.

lib/Solver/Solver.cpp
ChristianKaltenecker: This case should only happen if new operators in constraints are supported.
Should we throw an exception of some kind here?

lib/Solver/Solver.cpp
ChristianKaltenecker: Thanks; I will do that in the next revision.

lib/Solver/Solver.cpp
ChristianKaltenecker: That's right. That's a bit too much. I will change that.

lib/Solver/Solver.cpp
ChristianKaltenecker: All right; I will fix it.

include/vara/Solver/Error.h
ChristianKaltenecker: It's the same as in [Error.h in the feature folder](https://github.com/se-sic/vara-feature/blob/vara-dev/include/vara/Feature/Error.h). Maybe @s9latimm could help?

include/vara/Solver/ConfigurationFactory.h
ChristianKaltenecker: Will do that with the next revision.

include/vara/Solver/ConfigurationFactory.h
ChristianKaltenecker: Will do that in the next revision.

include/vara/Solver/ConfigurationFactory.h
ChristianKaltenecker: Hm, yeah... This will be changed too into one single iterator.

lib/Solver/Solver.cpp
vulder: We have two options here: a) remove the default case, normally we should get a warning if there is no default an a case is not handled (if not we could enable that warning) b) we go the less easily detectable route and just add an assert before the return, so that at least in debug builds we would find the case

lib/Solver/Solver.cpp
ChristianKaltenecker: What would be a good alternative here? I can't use uint64_t since it is in conflict with the error code (which is of the same type).

ChristianKaltenecker: The problem is that the compiler can then no longer distinguish whether I'm returning an error code or the actual number of configurations.

include/vara/Solver/SolverFactory.h
```diff
@@ -0,0 +1,63 @@
...
+/// solver using a certain feature model.
+class SolverFactory {
+public:
+  /// This method returns a pointer to an initialized solver. This solver has
+  /// already processed the features and constraints of the given feature model.
+
+  /// \param Model the model to use for the initialization of the solver
+  /// \param Type the type of solver to use
+
+  /// \returns a unique pointer containing the initialized solver
```
vulder:
```suggestion
  /// This method returns a pointer to an initialized solver. This solver has
  /// already processed the features and constraints of the given feature model.
  ///
  /// \param Model the model to use for the initialization of the solver
  /// \param Type the type of solver to use
  ///
  /// \returns a unique pointer containing the initialized solver
```

include/vara/Solver/SolverFactory.h
```diff
@@ -0,0 +1,63 @@
...
+  initializeSolver(const SolverType Type) {
+    vara::feature::FeatureModelBuilder B;
+    std::unique_ptr<const feature::FeatureModel> FM = B.buildFeatureModel();
+    return initializeSolver(*FM, Type);
+  }
+
+private:
+  /// This method returns an initialized Z3 solver.
+  /// \param Model the model to use for the initialization of the solver
+  /// \return a unique pointer containing the initialized solver
```
vulder:
```suggestion
  /// This method returns an initialized Z3 solver.
  ///
  /// \param Model the model to use for the initialization of the solver
  ///
  /// \return a unique pointer containing the initialized solver
```

lib/Solver/Z3Solver.cpp
```diff
@@ -87,7 +87,7 @@ Result<SolverErrorCode> Z3Solver::addFeature(const string &FeatureName) {
       OptionToVariableMapping.end()) {
     return ALREADY_PRESENT;
   }
-  z3::expr Feature = Context.bool_const(FeatureName.c_str());
+  z3::expr const Feature = Context.bool_const(FeatureName.c_str());
```
vulder:
```suggestion
  const z3::expr Feature = Context.bool_const(FeatureName.c_str());
```
We use west `const` in the code base

ChristianKaltenecker: Thanks for the info. I will change this in my code base. :+1:

lib/Solver/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+set(SOLVER_LIB_SRC
+        Z3Solver.cpp
```
vulder:
```suggestion
  Z3Solver.cpp
```

unittests/Solver/SolverFactory.cpp
```diff
@@ -0,0 +1,63 @@
...
+  B.makeRoot("root");
+  B.makeFeature<vara::feature::BinaryFeature>("Foo", true);
+  B.addEdge("root", "Foo");
+  B.makeFeature<vara::feature::BinaryFeature>("alt", false);
+  B.addEdge("root", "alt");
+  B.makeFeature<feature::BinaryFeature>("a", true);
+  B.makeFeature<feature::BinaryFeature>("b", true);
+  B.addEdge("alt", "a");
+  B.addEdge("alt", "b");
+  std::unique_ptr<feature::FeatureModel::BooleanConstraint> C =
```
vulder: jj

vulder (approved):

vulder (approved):

PR #93, opened 2022-08-23
Member Attribute With Index

s9latimm:
Related https://github.com/se-sic/vara-llvm-project/pull/64
Related https://github.com/se-sic/VaRA/pull/928

### DIFF
--- a/include/vara/Feature/FeatureSourceRange.h
+++ b/include/vara/Feature/FeatureSourceRange.h
@@ -101,6 +101,8 @@ public:
       return StrS.str();
     }
 
+    [[nodiscard]] size_t nested() const { return Class.size(); }
+
     [[nodiscard]] std::string memberName() const { return Member; }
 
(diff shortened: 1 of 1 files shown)


include/vara/Feature/FeatureSourceRange.h
```diff
@@ -101,6 +101,8 @@ class FeatureSourceRange {
       return StrS.str();
     }
 
+    [[nodiscard]] size_t nested() const { return Class.size(); }
```
vulder: Hmm, `nested` sounds like a `true/false` kind check. Maybe something like `nestingDepth`?

vulder (changes requested):

include/vara/Feature/FeatureSourceRange.h
vulder: And could you please write a small method doc to the function that explains what is returned.

vulder (approved):

PR #107, opened 2023-01-19
Support Non-Boolean and Mixed Contraints

s9latimm:
- [x] Implement constraint types (boolean, non-boolean, mixed)
- [x] Update parsing / builder / transaction
- [x] Update FM-Writer
- [x] Add new unit / integration tests for constraints
- [x] Pass existing tests / CI


include/vara/Feature/FeatureModel.h
```diff
@@ -34,8 +34,6 @@ class FeatureModel {
 
 public:
   using FeatureMapTy = llvm::StringMap<std::unique_ptr<Feature>>;
-  using ConstraintTy = Constraint;
-  using ConstraintContainerTy = std::vector<std::unique_ptr<ConstraintTy>>;
   using RelationshipContainerTy = std::vector<std::unique_ptr<Relationship>>;
```
vulder, line 37: Ok, maybe I don't get the design, but why is constraint gone? But a class added below?

vulder (changes requested):
Could you please add a bit of documentation so it's clearer what what does and how stuff works.

vulder (approved):

PR #108, opened 2023-01-31
Refactors llvm::Optional to std::optional

vulder:
As LLVM is currently in the process of removing llvm::Optional with 17, we should already start to switch to std::optional.


ChristianKaltenecker (approved):
Looks good to me. I will consider using std::optional in the other PR.

PR #109, opened 2023-02-03
Add Constraint Builder

s9latimm:
- [x] Implementation
  - [x] Precedence 
- [x] Documentation
- [x] Tests
- [x] Python bindings

Closes https://github.com/se-sic/VaRA/issues/702

### DIFF
--- a/include/vara/Feature/ConstraintBuilder.h
+++ b/include/vara/Feature/ConstraintBuilder.h
@@ -1,0 +1,241 @@
+#ifndef VARA_FEATURE_CONSTRAINTBUILDER_H
+#define VARA_FEATURE_CONSTRAINTBUILDER_H
+
+#include "vara/Feature/Constraint.h"
+#include "vara/Feature/Feature.h"
+
+namespace vara::feature {
+
+//===----------------------------------------------------------------------===//
+//                              ConstraintBuilder
+//===----------------------------------------------------------------------===//
+
+class ConstraintBuilder {
+public:
+  ConstraintBuilder() : Root(nullptr), Head(&Root) {}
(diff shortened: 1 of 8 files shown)


s9latimm:
@sarvech76

include/vara/Feature/ConstraintBuilder.h
```diff
@@ -0,0 +1,342 @@
...
+    }
+    Head = &Root;
+    return *this;
+  }
+
+  /// Open parenthesis. Allows to add more precise parentheses.
+  ///
+  /// Example:
+  ///     left().feature("A").implies().feature("B").right()
+  ///      .implies().feature("C").implies().feature("D")
```
vulder: Why does this example have not left/right/ for the right side `(C => D)`

vulder (approved):
looks every nice overall, we can integrate this after fixing the few things discussed on friday.

nice work :)

include/vara/Feature/ConstraintBuilder.h
s9latimm: Using the builder all operators are right-associative by default and parentheses inserted respectively, so we do not need to specify those here. It would be fine if one would do so, but not necessary.

vulder: How would you, with this syntax then express this expression: `((A => B) => C => D)` and  `(((A => B) => C) => D)`

s9latimm: You would need too explicitly parenthesize both expressions or simply use the `()` operator.

`feature("A").implies().feature("B")().implies().feature("C")().implies().feature("D")`

vulder (approved):

PR #115, opened 2023-03-18
Add a configuration generator

ChristianKaltenecker:
Adds a new tool called `config-generator`.
This tool exposes the `SampleSetParser` and `Solver` for either importing a sample set or generating all configurations for a given feature model.
This PR addresses se-sic/VaRA#838 (however, sampling strategies are missing; maybe open another Issue only for incorporating sampling strategies).


tools/config-generator/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+  LINK_PRIVATE
+  VaRAFeature
+  VaRASampling
+  VaRASolver
+  csv
+  ${STD_FS_LIB}
+)
+
+add_custom_target(
+  check-vara-config-generator
```
vulder, line 16: You still need to add this test target to the github CI workflow

tools/config-generator/ConfigurationGenerator.cpp
```diff
@@ -0,0 +1,118 @@
...
+        llvm::cl::cat(ConfigCreatorCategory));
+
+static llvm::cl::opt<std::string>
+    CsvInputFilePath("csv", llvm::cl::desc("Path to the csv input file."),
+                     llvm::cl::value_desc("filename"), llvm::cl::init(""),
+                     llvm::cl::cat(ConfigCreatorCategory));
+
+static llvm::cl::opt<std::string>
+    OutputFilePath("out", llvm::cl::desc("Path to the yml output file."),
+                   llvm::cl::value_desc("filename"), llvm::cl::init(""),
```
vulder:
```suggestion
                   llvm::cl::value_desc("filename"), llvm::cl::init("configurations.yaml"),
```

tools/config-generator/ConfigurationGenerator.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "llvm/Support/CommandLine.h"
+#include "llvm/Support/Errc.h"
+#include "llvm/Support/InitLLVM.h"
+#include "llvm/Support/MemoryBuffer.h"
+
+static llvm::cl::OptionCategory
+    ConfigCreatorCategory("Configuration generator options");
+
+static llvm::cl::list<std::string>
+    FileNames(llvm::cl::Positional, llvm::cl::desc("Path to the feature model"),
```
vulder: Why allow multiple file names if we only process one? We could just change this to a singe filename or?

tools/config-generator/ConfigurationGenerator.cpp
```diff
@@ -0,0 +1,118 @@
...
+    llvm::errs() << "error: Sampling is not implemented yet.\n";
+    return 1;
+  }
+
+  if (!OutputFilePath.empty() && !Configurations.empty()) {
+    const std::string Str =
+        vara::sampling::SampleSetWriter::writeConfigurations(*FM,
+                                                             Configurations);
+    std::error_code EC;
+    auto *Out = new llvm::raw_fd_ostream(OutputFilePath.getValue(), EC);
```
vulder: Why not just use a value here? It's not really necessary to heap allocate the ostream here.

vulder (changes requested):
A few nits but overall this looks good. Could you please also link the issue in the PR desc. so I know which issue to close :)

tools/config-generator/ConfigurationGenerator.cpp
ChristianKaltenecker: Yep. Changed it.

tools/config-generator/ConfigurationGenerator.cpp
ChristianKaltenecker: True; I changed this.

vulder (approved):

PR #125, opened 2023-09-15
Add python bindings for configurations

danjujan:


setup.py
```diff
@@ -47,6 +47,8 @@ def build_extension(self, ext):
             '-DPYTHON_EXECUTABLE=' + sys.executable,
             '-DVARA_FEATURE_USE_Z3_SOLVER=True'
         ]
+        if os.path.exists('/lib/x86_64-linux-gnu/libz3.so'):
+            cmake_args.append('-DVARA_FEATURE_BUILD_Z3_SOLVER=False')
```
vulder, line 51: I'm not sure if this is a good way to determine if we should build the solver.
1) the path is a system specific
2) we might want to build the solver from scratch even for those system, as we depend on a very specific version of z3. (The debian 11 can cause problems)

Maybe we should keep building the solver as the default and give the user a option to disable that? Or, if it works for you, only you use the local one.

vulder (changes requested):
Except for the comment below, LGTM

include/vara/Solver/Solver.h
```diff
@@ -131,6 +131,16 @@ class Solver {
...
+
+  // TODO (@boehmseb): Was this initial that there is not a decl. in the class?
+  // ++ This is not part of the Solver interface, which seems also strange
+  // (wrong)
+
+  virtual Result<SolverErrorCode,
+                 std::vector<std::unique_ptr<vara::feature::Configuration>>>
+  getAllValidConfigurations() = 0;
+
+  virtual Result<SolverErrorCode, uint64_t> getNumberValidConfigurations() = 0;
```
boehmseb: This part was removed by 864ff022ee and should stay removed.
The problem here is that these functions modify solver state in a way that interferes with the result of other solver functions.

I think this PR merged an older version of the branch that eventually was merged as 864ff022ee and therefore contains some outdated code. I try to mark all the relevant locations, but as far as I can see, all changes to the solver code in this PR should be reverted.

```suggestion
```

include/vara/Solver/Z3Solver.h
```diff
@@ -63,6 +63,12 @@ class Z3Solver : public Solver {
   Result<SolverErrorCode, std::unique_ptr<vara::feature::Configuration>>
   getNextConfiguration() override;
 
+  Result<SolverErrorCode,
+         std::vector<std::unique_ptr<vara::feature::Configuration>>>
+  getAllValidConfigurations() override;
+
+  Result<SolverErrorCode, uint64_t> getNumberValidConfigurations() override;
+
```
boehmseb:
```suggestion
```

include/vara/Solver/Z3Solver.h
```diff
@@ -91,6 +97,12 @@ class Z3Solver : public Solver {
 
   /// The current model of the SAT solver.
   std::optional<z3::model> CurrentModel;
+
+  /// Flag that indicates whether the solver state has been modified by calling
+  /// \c getNextConfiguration.
+  /// This is important for functions that want to enumerate all configurations,
+  /// like \c getAllValidConfigurations or \c getNumberValidConfigurations.
+  bool Dirty = false;
```
boehmseb:
```suggestion
```

lib/Solver/Z3Solver.cpp
```diff
@@ -183,14 +183,50 @@ Result<SolverErrorCode, bool> Z3Solver::hasValidConfigurations() {
 
 Result<SolverErrorCode, std::unique_ptr<vara::feature::Configuration>>
 Z3Solver::getNextConfiguration() {
+  if (Dirty) {
+    excludeCurrentConfiguration();
+  } else {
+    Dirty = true;
+  }
+
```
boehmseb:
```suggestion
```

lib/Solver/Z3Solver.cpp
```diff
@@ -183,14 +183,50 @@ Result<SolverErrorCode, bool> Z3Solver::hasValidConfigurations() {
...
   if (Solver->check() == z3::unsat) {
     return UNSAT;
   }
-  CurrentModel = Solver->get_model();
-  excludeCurrentConfiguration();
```
boehmseb: Need to be reinstated.

lib/Solver/Z3Solver.cpp
```diff
@@ -183,14 +183,50 @@ Result<SolverErrorCode, bool> Z3Solver::hasValidConfigurations() {
...
+  Solver->push();
+  auto Vector = std::vector<std::unique_ptr<vara::feature::Configuration>>();
+  while (auto Config = getNextConfiguration()) {
+    Vector.insert(Vector.begin(), Config.extractValue());
+  }
+  Solver->pop();
+  Dirty = false;
+  return Vector;
+}
+
```
boehmseb:
```suggestion
```

boehmseb (changes requested):

vulder (approved):

boehmseb (approved):

PR #126, opened 2023-10-06
Fix configuration generation

danjujan:
Unittest for https://github.com/se-sic/VaRA/issues/1043


vulder:
@ChristianKaltenecker Do you have an idea what is wrong with the solver here?

lib/Solver/Z3Solver.cpp
```diff
@@ -177,35 +177,44 @@ Result<SolverErrorCode, bool> Z3Solver::hasValidConfigurations() {
 
 Result<SolverErrorCode, std::unique_ptr<vara::feature::Configuration>>
 Z3Solver::getNextConfiguration() {
-  // Add previous configuration as a constraint
-  excludeCurrentConfiguration();
+  if (!Dirty) {
+    Dirty = true;
+  } else {
+    excludeCurrentConfiguration();
+  }
```
vulder:
```suggestion
  if (Dirty) {
    excludeCurrentConfiguration();
  } else {
    Dirty = true;
  }
```

lib/Solver/Z3Solver.cpp
```diff
@@ -258,11 +266,10 @@ Z3Solver::getCurrentConfiguration() {
   const z3::model M = Solver->get_model();
   auto Config = std::make_unique<vara::feature::Configuration>();
 
-  for (auto Iterator = OptionToVariableMapping.begin();
-       Iterator != OptionToVariableMapping.end(); Iterator++) {
-    const z3::expr OptionExpr = *Iterator->getValue();
+  for (auto const &Entry : OptionToVariableMapping) {
```
vulder:
```suggestion
  for (const auto &Entry : OptionToVariableMapping) {
```

lib/Solver/Z3Solver.cpp
```diff
@@ -232,18 +241,17 @@ Result<SolverErrorCode> Z3Solver::excludeCurrentConfiguration() {
   }
   const z3::model M = Solver->get_model();
   z3::expr Expr = Context.bool_val(false);
-  for (auto Iterator = OptionToVariableMapping.begin();
-       Iterator != OptionToVariableMapping.end(); Iterator++) {
-    const z3::expr OptionExpr = *Iterator->getValue();
+  for (auto const &Entry : OptionToVariableMapping) {
```
vulder:
```suggestion
  for (const auto &Entry : OptionToVariableMapping) {
```

vulder (changes requested):

boehmseb:
This issue is more complicated than assumed. The current state of this PR should be considered a quick-fix. Do not merge until it is properly resolved.

boehmseb:
This is now ready for review.

include/vara/Solver/ConfigurationFactory.h
```diff
@@ -17,7 +17,7 @@ class ConfigurationIterable {
 public:
   class ConfigurationIterator {
   public:
-    using iterator_category = std::forward_iterator_tag;
+    using iterator_category = std::input_iterator_tag;
```
vulder, line 20: Why can the config iterator no longer be a forward_iterator? Having it at least forward would probably be good to use it in standard algorithms.

boehmseb: As far as I understood this, a `forward_iterator` allows to iterate multiple times.
That was never possible with the current implementation because incrementing the iterator changes the solver state in a non-reversible way. So it was incorrect to use `forward_iterator` in the first place.

boehmseb: At least the following requirements of a `forward_iterator` are violated:
Given `a` and `b`, dereferenceable iterators of type `It`:
- If `a` and `b` compare equal (`a == b` is contextually convertible to `true`) then either they are both non-dereferenceable or `*a` and `*b` are references bound to the same object.
- `a == b` implies `++a == ++b`

For the first point, it would only be true that `*a` and `*b` describe the same configuration, but they are not the same object (they are  `unique_ptr`s pointing to a `Configuration`). In fact, such an object is created every time the iterator is dereferenced or incremented.

boehmseb: Thinking more about this, I'm not sure if this could also be an issue for the `input_iterator` specification as it says:

If `a` == `b` and `(a, b)` is in the domain of `==` then `*a` is equivalent to `*b`.

This would only be fulfilled if _equivalent_ refers to conceptual equivalence of the objects (i.e., they represent the same ting/configuration). There is currently no equivalence implemented on `Configuration`s.

vulder: Hmm, I see. Then let's keep the inputput tag. PS: I don't think they need to be the same object, so input_iterator should be fine.

vulder:
resolves https://github.com/se-sic/VaRA/issues/1043

vulder (approved):

PR #140, opened 2025-04-10
Support mac os m architecture

Kallistos:


CMakeLists.txt
```diff
@@ -20,9 +20,35 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
 
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-  add_compile_options(-stdlib=libstdc++)
+  if(APPLE)
+    # Clang on macOS must use libc++
+    add_compile_options(-stdlib=libc++)
+  elseif(UNIX AND NOT APPLE)
+    # Clang on Linux can use libstdc++ (usually default) No need to explicitly
+    # add this unless required
+    add_compile_options(-stdlib=libstdc++)
```
vulder, line 29: Could you make this optional or maybe don't set it at all, as it is currently not necessary to be specified for UNIX systems where the system default gets chosen.

PS: I compile with libc++ on some UNIX setup :smiling_face_with_tear:

CMakeLists.txt
```diff
@@ -20,9 +20,35 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
+  endif()
+endif()
+
+# if(APPLE) execute_process( COMMAND xcrun --show-sdk-path OUTPUT_VARIABLE
+# MACOS_SDK_PATH OUTPUT_STRIP_TRAILING_WHITESPACE )
+# include_directories("${MACOS_SDK_PATH}/usr/include/c++/v1") endif()
```
vulder:
```suggestion
```

Remove?

CMakeLists.txt
```diff
@@ -20,9 +20,35 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
+        CACHE STRING "" FORCE
+    )
+  endif()
 endif()
 
+if(APPLE)
+  set(Python3_ROOT_DIR "/opt/homebrew") # cmake-lint: disable=C0103
+endif()
+
+find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
```
vulder, line 46: Is this always necessary here? If I remember correctly, we only searched for python where we needed it for the bindings but I'm not sure if we did this our selfs or if we just used what pybind found.

CMakeLists.txt
```diff
@@ -20,9 +20,35 @@ set(CMAKE_CXX_EXTENSIONS OFF)
...
+  if(NOT DEFINED CMAKE_OSX_ARCHITECTURES)
+    set(CMAKE_OSX_ARCHITECTURES
+        "arm64"
+        CACHE STRING "" FORCE
+    )
+  endif()
 endif()
 
+if(APPLE)
+  set(Python3_ROOT_DIR "/opt/homebrew") # cmake-lint: disable=C0103
```
vulder, line 43: Oh boy, someone should fix the cmake find package for python :laughing: or why is it necessary to specify the actual root beforehand where it's not on other systems?

vulder (changes requested):

CMakeLists.txt
Kallistos: Yes, I do not need it, was not sure if that should be in it.
Should I remove it or what would be the way to make it optional to be helpful for Linux?

vulder: There are two options, either remove the line for unix and only keep it for apple or introduce a cmake options that let's one "choose" one of the two libs. something like `VARA_FEATURE_USE_STD_LIB` where one then can insert the lib wanted.

CMakeLists.txt
vulder: ?

